PR #4, opened 2017-05-23
Implementing exponential backoff alghoritm

### LINKED ISSUE #1: Exponential backoff
It would be nice if the client API supported an exponential backoff algorithm with a random wait, to automatically handle rate limiting of the server.

mtrunkat:
fixes #1


src/apify-error.js
```diff
@@ -0,0 +1,16 @@
...
+export const REQUEST_FAILED_ERROR_MESSAGE = 'Server request failed.';
+
+export default class ApifyError extends Error {
+    constructor(type, message, details) {
+        super(message);
+        this.name = APIFY_ERROR_NAME;
+        this.type = type;
+        this.details = details;
+
+        Error.captureStackTrace(this, ApifyError);
```
mtrunkat: 1) this way we don't display ApifyError constructor in error stack

2) If we don't set name using `this.name = APIFY_ERROR_NAME` then there will be just `Error` instead of `ApifyError` displayed in error

src/utils.js
```diff
@@ -1,39 +1,106 @@
 import request from 'request';
 import _ from 'underscore';
+import ApifyError, {
+    INVALID_PARAMETER_ERROR_TYPE,
+    REQUEST_FAILED_ERROR_TYPE,
+    REQUEST_FAILED_ERROR_MESSAGE,
+} from './apify-error';
+
+const RATE_LIMIT_EXCEEDED_STATUS_CODE = 429;
+const EXP_BACKOFF_MILLIS = 500;
```
mtrunkat: > https://github.com/Apifier/apify-client-js/issues/8

test/apify-error.js
```diff
@@ -0,0 +1,15 @@
+import { expect } from 'chai';
+import ApifyError, { APIFY_ERROR_NAME } from '../build/apify-error';
+
+describe('ApifyError', () => {
+    it('should correctly handle all the informations', () => {
```
mtrunkat: Fixed

PR #11, opened 2017-05-24
Simplyfing configuration to only baseUrl instead of port/protocol/hos…

### LINKED ISSUE #9: Change API configuration of base url from separate port/host/protocol/... to simply baseUrl

mtrunkat:
fixes #9

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -67,11 +67,11 @@ const ApifyClient = function (options = {}) {
         return (callOpts, callback) => {
             const mergedOpts = Object.assign({}, instanceOpts, callOpts);
 
-            // Remove traling forward slash from baseUrl.
-            if (mergedOpts.baseUrl.substr(-1) === '/') mergedOpts.baseUrl = mergedOpts.baseUrl.slice(0, -1);
-
             // Check that all required parameters are set.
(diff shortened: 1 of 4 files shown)


mtrunkat:
TODO: remove trailing forward slash

src/index.js
```diff
@@ -68,22 +67,11 @@ const ApifyClient = function (options = {}) {
         return (callOpts, callback) => {
             const mergedOpts = Object.assign({}, instanceOpts, callOpts);
 
+            // Remove traling forward slash from baseUrl.
+            if (mergedOpts.baseUrl.substr(-1) === '/') mergedOpts.baseUrl = mergedOpts.baseUrl.slice(0, -1);
```
jancurn: Shouldn't this be `instanceOpts.baseUrl` instead? and have this after the `if (!instanceOpts.baseUrl)` check?

mtrunkat: it's because user can override every parameter in method call also

apifyClient.crawlers.getCrawler({ crawlerId: 'ljknjk', baseUrl: 'urlWithTrailingSlash/' )

jancurn: ok just beware that `mergedOpts.baseUrl` might be null so this will throw an error, maybe the below check should be earlier and of form `if (! mergedOpts.baseUrl)`

PR #15, opened 2017-05-26
Feature/input check

mtrunkat:


src/key_value_stores.js
```diff
@@ -1,69 +1,115 @@
...
-            'Content-Type': contentType,
-        },
-    }),
-
-    deleteRecord: (requestPromise, { baseUrl, storeId, recordKey }) => requestPromise({
-        url: `${baseUrl}${BASE_PATH}/${storeId}/records/${recordKey}`,
-        json: true,
-        method: 'DELETE',
-    }),
+    putRecord: (requestPromise, { baseUrl, storeId, recordKey, body, contentType = 'text/plain' }) => {
```
jancurn: Wouldn't be better and nicer to use just `key` instead of `recordKey` ?

src/utils.js
```diff
@@ -112,3 +114,20 @@ export const requestPromise = (options, iteration = 0) => {
...
+/**
+ * Checks that given parameter is of given type and throws ApifyError.
+ * If errorMessage is not provided then error message is created from name and type of param.
+ *
+ * @param type         string - "String", "Number", ... (see ee: https://github.com/gkz/type-check)
+ * @param value        string - user entered value of that parameter
+ * @param name         string - parameter name (crawlerId for options.crawlerId)
+ * @param errorMessage string - optional error message
+ */
+export const checkParamOrThrow = (type, value, name, errorMessage) => {
```
jancurn: I'd switch the order of `type` and `value` parameters, it easier to follow in the code, especially if the pattern is longer. Also the name `checkParamOrThrow` talks about "param", so the param should have a prominent position.

lubos-turek: @jancurn I think it's good to have parameters `value` and `name` next to each other as they "look" so similar.

`checkParamOrThrow('String', baseUrl, 'baseUrl');` is better than `checkParamOrThrow(baseUrl, String, 'baseUrl');`

In the second case it would be easy to modify one and forget about the other one and get wrong error messages.

It's a minor thing, just my 2 cents...

src/key_value_stores.js
mtrunkat: done

src/utils.js
mtrunkat: @lubos-turek @jancurn - so what about - value, name, type ?

lubos-turek: fine with me :+1:

PR #16, opened 2017-05-29
Feature/crawler endpoints

### LINKED ISSUE #10: Cover all crawler endpoints
Client still does not cover these endpoints:

1. Create crawler
2. Get crawler settings
3. Update crawler settings
4. Delete crawler
5. Stop execution
6. Get list of executions
7. Get last execution
8. Get last execution results

\+ proper unit tests

lubos-turek:
Fixes #10

### DIFF
--- a/src/crawlers.js
+++ b/src/crawlers.js
@@ -130,7 +130,7 @@ export default {
         });
     },
 
-    startCrawler: (requestPromise, options) => {
+    startExecution: (requestPromise, options) => {
         const { crawler, token, userId } = options;
 
         checkParamOrThrow(userId, 'userId', 'String');
--- a/src/crawlers.js
+++ b/src/crawlers.js
@@ -3,42 +3,6 @@ import _ from 'underscore';
 import { checkParamOrThrow } from './utils';
 
 export const BASE_PATH = '/v1';
-export const CRAWLER_ATTRIBUTES = [
-    'customId',
-    '_id',
-    'comments',
-    'startUrls',
-    'crawlPurls',
-    'pageFunction',
-    'clickableElementsSelector',
-    'interceptRequest',
-    'considerUrlFragment',
-    'loadImages',
-    'loadCss',
-    'injectJQuery',
-    'injectUnderscoreJs',
-    'ignoreRobotsTxt',
-    'skipLoadingFrames',
-    'verboseLog',
-    'disableWebSecurity',
-    'maxCrawledPages',
-    'maxOutputPages',
-    'maxCrawlDepth',
-    'timeout',
-    'resourceTimeout',
-    'pageLoadTimeout',
-    'pageFunctionTimeout',
-    'maxInfiniteScrollHeight',
-    'randomWaitBetweenRequests',
-    'maxCrawledPagesPerSlave',
-    'maxParallelRequests',
-    'customHttpHeaders',
-    'customProxies',
-    'cookies',
-    'cookiesPersistence',
-    'customData',
-    'finishWebhookUrl',
-];
 
 export default {
     listCrawlers: (requestPromise, options) => {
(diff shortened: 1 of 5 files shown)


lubos-turek:
Fixes #10

mtrunkat:
```
apify.crawlers.startCrawler({
    crawlerId,
    settings: {

    }
})
```

src/crawlers.js
```diff
@@ -81,25 +153,75 @@ export default {
...
+
+        return requestPromise({
+            url: `${options.baseUrl}${BASE_PATH}/${userId}/crawlers/${crawler}/lastExec`,
+            json: true,
+            method: 'GET',
+            qs: queryString,
+        });
+    },
+
     getExecutionResults: (requestPromise, options) => {
```
mtrunkat: This method is missing pagination/format/... parameters.

lubos-turek: caused by a misleading diff. Whole file: https://github.com/Apifier/apify-client-js/blob/8ab604994ff8749b0d4f726a5b56f82163a80a58/src/crawlers.js#L218

src/crawlers.js
```diff
@@ -1,105 +1,187 @@
...
-        }
-        if (!options.token) {
-            throw new Error('Missing required parameter: token');
-        }
+        const { userId, token } = options;
+
+        checkParamOrThrow(userId, 'userId', 'String');
+        checkParamOrThrow(token, 'token', 'String');
 
         const queryString = _.pick(options, 'token', 'offset', 'query', 'desc');
```
mtrunkat: There should be `limit` instead of `query`.

src/crawlers.js
```diff
@@ -1,105 +1,187 @@
...
 
         return requestPromise({
```
mtrunkat: We should encapsulate returned array in `items` parameter for future compatibility with pagination implementation:

```
{ 
    items: [ crawler1, crawler2, ... ],
}
```

Because we will have to add info about pagination which is received in headers - see https://www.apifier.com/api-reference#/introduction/pagination

So we will be then returning something like:

```
{
    items: [ crawler1, crawler2, ... ],
    count: 20,
    totalCount: 50,
    ...
}

File: src/crawlers.js [hunk 4]
```diff
@@ -1,105 +1,187 @@
...
+            qs: { token },
+        });
+    },
+
+    startExecution: (requestPromise, options) => {
+        const { crawlerId, userId, token, settings } = options;
+
+        checkParamOrThrow(userId, 'userId', 'String');
+        checkParamOrThrow(token, 'token', 'String');
+        checkParamOrThrow(crawlerId, 'crawlerId', 'String');
```
mtrunkat: We can add validation of settings as

```
checkParamOrThrow(settings, 'settings', 'Maybe Object');
```

because for `_.isEmpty` returns `false` for `string` which is invalid.

src/crawlers.js
```diff
@@ -1,105 +1,187 @@
...
+            json: true,
+            method: 'POST',
+            url: `${options.baseUrl}${BASE_PATH}/execs/${executionId}/stop`,
+            qs: { token },
+        };
+
+        return requestPromise(requestParams);
+    },
+
+    getListOfExecutions: (requestPromise, options) => {
```
mtrunkat: Same as `getListOfCrawlers` - we should encapsulate result in `{ items: [...] }`

src/crawlers.js
```diff
@@ -111,6 +193,26 @@ export default {
         return requestPromise(requestParams);
     },
 
+    getLastExecutionResults: (requestPromise, options) => {
```
mtrunkat: Same as `getListOfCrawlers` - we should encapsulate result in `{ items: [...] }` or something like that

src/crawlers.js
lubos-turek: done https://github.com/Apifier/apify-client-js/blob/7e69527d6786d4c4d520229b56c2b5f7de785930/src/crawlers.js#L111

src/crawlers.js
lubos-turek: done as discussed

PR #33, opened 2017-08-09
Feature/api new options

mtrunkat:


src/acts.js
```diff
@@ -9,7 +9,9 @@ import { checkParamOrThrow, pluckData, catchNotFoundOrThrow, encodeBody } from '
  */
 
 export const BASE_PATH = '/v2/acts';
-export const WAIT_FOR_FINISH_SECS = 120;
+export const MAX_WAIT_FOR_FINISH_SECS = 120; // This is used in Apify.call()
```
jancurn: We don't really need this constant, in Apify.call() we can use waitForFinish=9999999, and in worst case the API will return after 120 seconds. If we change this value in API in the future, we won't need to update anything

PR #63, opened 2018-04-19
Fix doc for users method #62

drobnikj:


src/users.js
```diff
@@ -1,20 +1,26 @@
...
+     * Returns private and public information about user account.
      *
      * @memberof ApifyClient.users
      * @instance
      * @param {Object} options
-     * @param {String} [options.token]
-     * @param {String} [options.userId = 'me'] - User ID or username of the user
-     * @param callback
-     * @returns {UserInfo}<
+     * @param {String} [options.token] - If is set, it shows private and public information about user account, otherwise only public.
```
jancurn: "If set, the function returns a private and public information for the user account, otherwise it only returns public information."

src/users.js
```diff
@@ -1,20 +1,26 @@
...
+     * @param {String} [options.userId = 'me'] - User ID or username of the user,
+     *                                          if not set it returns information about current user account.
```
jancurn: "Desired user ID or username. By default it is `me`, which causes the function to return information about the current user."

PR #74, opened 2018-08-02
Add support for parsing HTML and XML

mnmkng:
Adds HTML and XML to `parseBody()`
to fix: https://github.com/apifytech/apify-js/issues/130
and support: https://github.com/apifytech/apify-js/pull/140

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -18,7 +18,9 @@ const RATE_LIMIT_EXCEEDED_STATUS_CODE = 429;
 const EXP_BACKOFF_MILLIS = 500;
 const EXP_BACKOFF_MAX_REPEATS = 8; // 64s
 const CONTENT_TYPE_JSON = 'application/json';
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -18,7 +18,9 @@ const RATE_LIMIT_EXCEEDED_STATUS_CODE = 429;
 const EXP_BACKOFF_MILLIS = 500;
 const EXP_BACKOFF_MAX_REPEATS = 8; // 64s
 const CONTENT_TYPE_JSON = 'application/json';
+const CONTENT_TYPE_XML = 'application/xml';
 const CONTENT_TYPE_TEXT_PLAIN = 'text/plain';
+const CONTENT_TYPE_TEXT_HTML = 'text/html';
```
mtrunkat: I guess that anything `text/*` could be parsed using `.toString()`

mtrunkat: But I am not completely sure, we may need to confirm this assumption

PR #83, opened 2018-11-01
Feature/async await

mtrunkat:


src/utils.js
```diff
@@ -65,6 +65,34 @@ export const newApifyClientErrorFromResponse = (statusCode, body, isApiV1) => {
...
 };
 
+/**
+ * Promisified version of request() package.
+ *
+ * @param  {String} method
+ * @param  {Object} options
+ * @return {Promise<[response, body]>}
+ */
+const promisifiedRequest = (method, options) => {
```
mnmkng: Why not use `request-promise-native` package?

mtrunkat: Yeah good point, was there before we supported selected Promise implementation from options.

PR #87, opened 2018-11-06
Feature/ap 526 actor versions

drobnikj:


mtrunkat:
Looks good. I would just rename `listActVersions` to `listActorVersions`, etc.

Because we will be renaming client.acts to client.actors

jancurn:
Agree with Marek, otherwise looks good to me.

PR #89, opened 2018-11-16
Allowing to overload actor task input and options

mtrunkat:


src/tasks.js
```diff
@@ -273,29 +273,48 @@ export default {
...
+        checkParamOrThrow(contentType, 'contentType', 'Maybe String');
+        checkParamOrThrow(timeout, 'timeout', 'Maybe Number');
+        checkParamOrThrow(memory, 'memory', 'Maybe Number');
+        checkParamOrThrow(build, 'build', 'Maybe String');
 
         const safeTaskId = replaceSlashWithTilde(taskId);
         const query = {};
 
         if (waitForFinish) query.waitForFinish = waitForFinish;
         if (token) query.token = token;
```
mnmkng: Since token is mandatory, there should be no need to check for its existence.

mtrunkat: You are right. It should not be mandatory as actor might be anonymously runnable!

PR #94, opened 2019-01-14
Feature/dataset hidden fields

mtrunkat:


src/datasets.js
```diff
@@ -247,7 +247,15 @@ export default {
      *   Overrides the default element name that wraps each page or page function result object in XML output.
      *   By default, the element name is `page` or `result`, depending on the value of the `simplified` option.
      * @param {Boolean} [options.skipHeaderRow]
-     *   If set to `1` then header row in csv format is skipped.
+     *   If set to `true` then header row in csv format is skipped.
+     * @param {Boolean} [options.clean]
+     *   If `true` then returns only nonempty items without hidden fields (fields starting with "#" character).
```
jancurn: If `true` then the function returns only non-empty items and skips hidden fields (i.e. fields starting with `#` character). Note that the `clean` parameter is a shortcut for `skipHidden: true` and `skipEmpty: true` options.

src/datasets.js
```diff
@@ -247,7 +247,15 @@ export default {
...
+     *   Note that in this case the returned number of items might be lower than limit parameter and pagination must be done using the `limit` value.
+     * @param {Boolean} [options.skipHidden]
+     *   If `true` then doesn't return hidden fields (fields starting with "#" character).
+     * @param {Boolean} [options.skipEmpty]
+     *   If `true` then doesn't return empty items.
```
jancurn: "then the function"

src/datasets.js
```diff
@@ -247,7 +247,15 @@ export default {
      *   Overrides the default element name that wraps each page or page function result object in XML output.
      *   By default, the element name is `page` or `result`, depending on the value of the `simplified` option.
      * @param {Boolean} [options.skipHeaderRow]
-     *   If set to `1` then header row in csv format is skipped.
+     *   If set to `true` then header row in csv format is skipped.
+     * @param {Boolean} [options.clean]
+     *   If `true` then returns only nonempty items without hidden fields (fields starting with "#" character).
+     *   Note that in this case the returned number of items might be lower than limit parameter and pagination must be done using the `limit` value.
+     * @param {Boolean} [options.skipHidden]
+     *   If `true` then doesn't return hidden fields (fields starting with "#" character).
```
jancurn: "then the function"

PR #95, opened 2019-01-21
Overriding toString of an ApifyClientError to inlcude details

mtrunkat:


src/apify_error.js
```diff
@@ -16,4 +18,18 @@ export default class ApifyClientError extends Error {
 
         Error.captureStackTrace(this, ApifyClientError);
     }
+
+    toString() {
+        let details = '';
+
+        if (this.details) {
+            const items = [];
+            _.mapObject(this.details, (val, key) => {
```
mnmkng: I guess I'd prefer something like:

```js
details = Object.entries(this.details)
    .map(([key, val]) => `${key}=${JSON.stringify(val)}`)
    .join(', ');
```

but that's probably just me not being very used to underscore.

src/apify_error.js
```diff
@@ -16,4 +18,18 @@ export default class ApifyClientError extends Error {
...
+                items.push(`${key}=${JSON.stringify(val)}`);
+            });
+            details = `(${items.join(', ')})`;
+        }
+
+        return `[${this.name}] ${APIFY_ERROR_NAME}: ${this.message} ${details}`;
```
mnmkng: Why no stack?

jancurn: good point, maybe we should include it too

PR #102, opened 2019-02-21
Feature - Exponential Backoff for requestPromise()

petrpatek:


src/utils.js
```diff
@@ -96,8 +86,8 @@ export const requestPromise = async (options, stats) => {
     const INVALID_PARAMETER_ERROR_TYPE = isApiV1 ? INVALID_PARAMETER_ERROR_TYPE_V1 : INVALID_PARAMETER_ERROR_TYPE_V2;
     const REQUEST_FAILED_ERROR_TYPE = isApiV1 ? REQUEST_FAILED_ERROR_TYPE_V1 : REQUEST_FAILED_ERROR_TYPE_V2;
 
-    const expBackOffMillis = options.expBackOffMillis || EXP_BACKOFF_MILLIS;
-    const expBackOffMaxRepeats = options.expBackOffMaxRepeats || EXP_BACKOFF_MAX_REPEATS;
+    const expBackoffMillis = options.expBackOffMillis || EXP_BACKOFF_MILLIS;
```
mnmkng: We should either use `Backoff` or `BackOff`, not both.

petrpatek: @mnmkng  It sounds reasonable. I should probably rename that in `apify-shared` since it will not break anything, even though `expBackoffMillis` is correct.

jancurn: "backoff" is correct, let's stick with that. "backOff" or "BACK_OFF" looks bad

petrpatek:
@jancurn According to the test coverage it looks good.

PR #103, opened 2019-02-21
Bug fix: Dataset Exponential Backoff

petrpatek:
I should probably add some tests that it really retries those errors. Is there a way how to stub the requestPromise function?

### DIFF
--- a/src/datasets.js
+++ b/src/datasets.js
@@ -1,6 +1,14 @@
 import _ from 'underscore';
 import { retryWithExpBackoff, RetryableError } from 'apify-shared/exponential_backoff';
-import { checkParamOrThrow, gzipPromise, pluckData, catchNotFoundOrThrow, parseBody, wrapArray, parseDateFields } from './utils';
+import {
+    checkParamOrThrow,
+    gzipPromise,
+    pluckData,
+    catchNotFoundOrThrow,
+    wrapArray,
+    parseDateFields,
+} from './utils'; // eslint-disable-line import/no-duplicates
+import * as Utils from './utils'; // eslint-disable-line import/no-duplicates
 
 
 /**
@@ -309,7 +317,6 @@ export default {
         if (options.skipEmpty) query.skipEmpty = 1;
 
         if (query.fields) query.fields = query.fields.join(',');
-
         const requestOpts = {
             url: `${baseUrl}${BASE_PATH}/${datasetId}/items`,
             method: 'GET',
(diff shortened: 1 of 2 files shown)


src/datasets.js
```diff
@@ -322,13 +324,25 @@ export default {
         const parseResponse = (response) => {
             const contentType = response.headers['content-type'];
             const wrappedItems = wrapArray(response);
-            if (!disableBodyParser) wrappedItems.items = parseBody(wrappedItems.items, contentType);
+            try {
+                if (!disableBodyParser) wrappedItems.items = parseBody(wrappedItems.items, contentType);
+            } catch (e) {
```
mtrunkat: We should not be catching any error here because other error that may potentially happen is that JSON is too large to be parsed and in that case we don't want to retry.

src/datasets.js
```diff
@@ -322,13 +324,25 @@ export default {
...
+                throw new RetryableError(e);
+            }
             return wrappedItems;
         };
 
-        return requestPromise(requestOpts)
-            .then(parseResponse)
-            .catch(catchNotFoundOrThrow);
+        const getData = async () => {
```
mtrunkat: Originally I wanted to write that those 3 lines `request().then().catch()` are faster to read than these 7 `let...; try {...} catch{...} return...` but then a realized that this will probably produce much better stack trace which is better

src/datasets.js
jancurn: But we should definitely handle the "Unexpected end of JSON input" error this way

src/datasets.js
mnmkng: When `requestPromise` throws, `response` will be `undefined` and `parseResponse` will throw that it cannot get property headers of undefined. @petrpatek 

We should:
```js
try {
    const response = await requestPromise(requestOpts);
    return parseResponse(response);
} catch (err) {
    catchNotFoundOrThrow(err);
}
```

mtrunkat: Yeah, you are right. catchNotFoundOrThrow() returns null in a case of 404 which should be returned.

mtrunkat: @petrpatek could you add test for this?

src/datasets.js
petrpatek: So I can catch this particular error based on its message and rethrow the Retryable error otherwise I will throw the original one. Am I right?

src/datasets.js
petrpatek: @mtrunkat Yes I will.

src/datasets.js
mtrunkat: Yes, exactly we want to solve "Unexpected end of JSON input" error only because in a case of other error retry won't help

src/datasets.js
```diff
@@ -373,3 +376,26 @@ export default {
...
     },
 };
+
+export function parseResponse(response, disableBodyParser) {
+    const contentType = response.headers['content-type'];
+    const wrappedItems = wrapArray(response);
+    try {
+        if (!disableBodyParser) wrappedItems.items = Utils.parseBody(wrappedItems.items, contentType);
+    } catch (e) {
+        if (e.message.includes('Unexpected end of JSON input')) {
```
jancurn: Please add some comment why this error is retryable

mtrunkat: And I would call it `parseDatasetItemsResponse` to make it clear

src/datasets.js
```diff
@@ -373,3 +376,26 @@ export default {
...
+            throw new RetryableError(e);
+        }
+        throw e;
+    }
+    return wrappedItems;
+}
+
+export async function getData(requestPromise, requestOpts, disableBodyParser) {
```
mtrunkat: `getDatasetItems()` ?

test/datasets.js
```diff
@@ -223,6 +227,84 @@ describe('Dataset', () => {
...
+                    headers: {
+                        'content-type': 'application/json',
+                    },
+                    body: {
+
+                    },
+                };
+                const stub = sinon.stub(utils, 'parseBody');
+                stub.callsFake(() => {
+                    throw new Error('Unexpected end of JSON input');
```
mtrunkat: Hey what about to create that error this way 👍 to ensure that has correct format

```
stub.callsFake(() => {
    JSON.parse('{"foo":')
});
```

PR #105, opened 2019-02-25
Adding metamorph

mtrunkat:


src/acts.js
```diff
@@ -387,6 +387,62 @@ export default {
...
+
+        checkParamOrThrow(baseUrl, 'baseUrl', 'String');
+        checkParamOrThrow(token, 'token', 'String');
+        checkParamOrThrow(actId, 'actId', 'String');
+        checkParamOrThrow(runId, 'runId', 'String');
+        checkParamOrThrow(targetActorId, 'targetActorId', 'String');
+        checkParamOrThrow(contentType, 'contentType', 'Maybe String');
+        checkParamOrThrow(build, 'build', 'Maybe String');
+
+        const safeActId = replaceSlashWithTilde(actId);
```
mnmkng: Perhaps use `actorId` and `safeActorId`.

We can do:

```js 
const { /* args */ actId: actorId /* args */ } = options;
```

And only replace this single assignment when we change the naming.

src/acts.js
```diff
@@ -387,6 +387,62 @@ export default {
             .then(parseDateFields);
     },
 
+    /**
+     * Runs the latest build of given act.
```
jancurn: Bad comment

PR #112, opened 2019-03-28
Fixed bug in rate limit stats

gippy:
I think the index in the array is wrong

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -120,7 +120,7 @@ export const requestPromise = async (options, stats) => {
 
         if (statusCode === RATE_LIMIT_EXCEEDED_STATUS_CODE && stats) {
             // Make sure this doesn't fail when someone increases number of retries on anything.
-            if (typeof stats.rateLimitErrors[iteration] === 'number') stats.rateLimitErrors[iteration - 1]++;
+            if (typeof stats.rateLimitErrors[iteration] === 'number') stats.rateLimitErrors[iteration]++;
             else stats.rateLimitErrors[iteration] = 1;
         }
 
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -120,7 +120,7 @@ export const requestPromise = async (options, stats) => {
 
         if (statusCode === RATE_LIMIT_EXCEEDED_STATUS_CODE && stats) {
             // Make sure this doesn't fail when someone increases number of retries on anything.
-            if (typeof stats.rateLimitErrors[iteration] === 'number') stats.rateLimitErrors[iteration - 1]++;
+            if (typeof stats.rateLimitErrors[iteration] === 'number') stats.rateLimitErrors[iteration]++;
```
jancurn: Actually it should be `iteration-1`, `iteration` is `1` in the first loop

gippy: Hmm you are right, then the other two index into the array are probably wrong. Since the condition checks `iteration`, but then sets `iteration-1`, same for the else branch where it sets `iteration` to 1

jancurn: So, how to fix this? Maybe @mnmkng can advice?

mnmkng: There needs to be `-1` everywhere I believe. I did not have the `if {} else {}` wrapper there initially and then I figured I'll throw it in "just in case", but apparently I did not update all the indices.

mtrunkat:
@gippy @mnmkng , can you take a look at this so that we can merge it?

Would be great to also fix this https://github.com/apifytech/apify-client-js/issues/110 as a part of this PR.

jancurn:
Good point by @mtrunkat, let's just push the items, no need to use index

PR #115, opened 2019-04-24
Adding clientId to request queue - create/update/delete operations

mtrunkat:


src/datasets.js
```diff
@@ -311,6 +311,7 @@ export default {
         if (options.skipHeaderRow) query.skipHeaderRow = 1;
         if (options.desc) query.desc = 1;
         if (options.bom) query.bom = 1;
+        else if (options.bom === false) query.bom = 0;
```
mnmkng: This seems like a really dirty fix of something. Why do we need this special behavior for `bom`?

jancurn: I agree, it looks quite bad, but I think we need to have it there. A comment would be good.

src/request_queues.js
```diff
@@ -227,24 +227,31 @@ export default {
      *                                        and to the end of the queue otherwise.
      * @param {String} [options.token] - Your API token at apify.com. This parameter is required
      *                                   only when using "username~queue-name" format for queueId.
+     * @param {String} [options.clientId] - Unique ID identifying client accessing the request queue.
```
jancurn: As mentioned in PR on BitBucket, I'd strongly prefer to call this `clientKey`, it's not really an ID.

src/datasets.js
mtrunkat: It's because default value of bom is `true`

mtrunkat: We should have called this variable `skipBom` to have it alligned

jancurn: Too late now. Actually, default value of `bom` is "depends on the format", for CSV it's true, for JSON it's not etc. Therefore `skipBom` wouldn't cut it either. Can we just add a comment explaining this?

mtrunkat: Comment updated

PR #132, opened 2019-12-05
Removing everything crawler related

mtrunkat:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+xxxxxxxxxxxxxxxxxxx
+===================
+- Removed legacy Apify Crawler methods that are no longer supported by Apify API.
```
jancurn: shall we call it breaking change?

mtrunkat: We should in principle but on the other hand, it doesn't work anyway

PR #138, opened 2020-02-17
Isomorphic client

petrpatek:
Just testing Github actions for now.


petrpatek:
## Breaking changes

- it is not possible to init without new
- no callbacks
- no default params such as `defaultDatasetId` in the client class - ids must be passed to the functions directly. @drobnikj I remember that it might be a problem, right? 
- all the backward compatible stuff from V1 API
- is that all @mnmkng?

## Things to discuss
 - test coverage
 - automation build/publish workflow for browsers (script tag, or npm module) and NodeJs(npm module).
- TODOs

test/mock_server/routes/actors.js
```diff
@@ -0,0 +1,54 @@
...
+                ? null
+                : { data: { id } };
+            res
+                .status(responseStatusCode)
+                .json(payload);
+        };
+    },
+};
+
+function addRoutes(router, routes) {
```
mnmkng: Should be factored out.

src/datasets.js
```diff
@@ -10,7 +10,8 @@ import {
 } from './utils'; // eslint-disable-line import/no-duplicates
 import * as Utils from './utils'; // eslint-disable-line import/no-duplicates
```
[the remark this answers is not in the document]
petrpatek: It is because of the stubs in tests. The two methods that are called  by `utils.*` are mocked the rest is not and I would find quite irritating to write all the `pluckData` with the `utils` prefix.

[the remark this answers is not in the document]
mnmkng: Could we have one import statement of `utils` and destructure from it below, rather than two import statements then?

src/datasets.js
```diff
@@ -75,37 +76,60 @@ import * as Utils from './utils'; // eslint-disable-line import/no-duplicates
...
 export const SIGNED_URL_UPLOAD_MIN_BYTESIZE = 1024 * 256;
 
-export default {
+export default class Datasets {
+    constructor(httpClient) {
+        this.basePath = '/v2/datasets';
+        this.client = httpClient;
+    }
+
+    _call(userOptions, endpointOptions) {
```
mtrunkat: This method is here 4 times. You could make a default abstract class and inherit from it

src/datasets.js
```diff
@@ -75,37 +76,60 @@ import * as Utils from './utils'; // eslint-disable-line import/no-duplicates
...
+        const callOptions = this._getCallOptions(userOptions, endpointOptions);
+        return this.client.call(callOptions);
+    }
+
+    _getCallOptions(userOptions, endpointOptions) {
```
mtrunkat: Also this one is copied all over

mtrunkat:
What about splitting the code into two dirs - one for helper stuff like `utils`, `http-client`, `error`, ... and another one for actual endpoint definitions.

src/http-client.js
```diff
@@ -0,0 +1,218 @@
...
+            // Override the default transform function to get unparsed response.
+            transformResponse: res => res,
+            httpAgent: this.httpAgent,
+            httpsAgent: this.httpsAgent,
+            validateStatus: null,
+            responseType: options.encoding === null ? 'arraybuffer' : null,
+        };
+    }
+
+    _getResponseLike(axiosResponse) {
```
mtrunkat: I'd add a comment here that one part handles server-side and the latter one browser response

src/http-client.js
```diff
@@ -0,0 +1,218 @@
...
+                const axiosResponse = await this.axiosInstance.request(axiosConfig); // eslint-disable-line
+                // TODO Emulate Request API for now
+                response = this._getResponseLike(axiosResponse);
+                statusCode = response ? response.statusCode : null;
+
+                // TODO Retest this with Axios
+                // It may happen that response is incomplete but request package silently returns original
+                // response as string instead of throwing an error. So we call JSON.parse() manually here.
+                // If parsing throws then the request gets retried with exponential backoff.
+                if (options.json && response.body) response.body = JSON.parse(response.body);
```
mtrunkat: Shouldn't we parse JSON this only when status code is < 300. 

Also keep in mind that API can redirect (for example reads from key-value store might be redirected to S3)

src/index.js
```diff
@@ -37,16 +35,17 @@ const getDefaultOptions = () => ({
...
-    keyValueStores,
-    datasets,
-    requestQueues,
-    logs,
-    users,
-    webhooks,
-    webhookDispatches,
+const endpointClasses = {
+    actors: Actors,
+    acts: Actors,
```
mtrunkat: Maybe we could add a decorator to `.acts.` methods that would print the deprecation warning.

src/http-client.js
mnmkng: I'm quite certain `axios` handles redirects before it even gives you access to the response.

src/index.js
mnmkng: The warnings are already there.

src/http-client.js
mtrunkat: Ok. But we are still parsing 4xx and 5xx response although it gets retried below.

src/index.js
mtrunkat: Warnings are there for old method names i.e.

```
client.acts.getAct();
client.actors.getAct();
```

but not for an old namespace with the new method name

```
client.acts.getActor();
```

src/http-client.js
mnmkng: Yeah. On the other hand, we use the `response.body` down on line 188. The whole piece of code is from the original client implementation with some very minor fixes to make it `axios` compatible. It needs another refactoring, but let's keep it out of this PR so that we can release a working browser build first and deal with the optimizations later.

jancurn:
Please let me know once this is in some more final state, will have a look.

src/index.js
mnmkng: Hmm, good point! Although those methods never existed before (in old namespace) so it feels weird to deprecate them, since nobody could be using them.

.babelrc
```diff
@@ -1,12 +1,9 @@
...
-      }
-    }]
-  ],
-  "plugins": [
-    "add-module-exports"
-  ]
+    "presets": [
+        ["@babel/preset-env", {
+            "targets": {
+                "node": "8.14.0"
```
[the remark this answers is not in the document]
petrpatek: Yes, we should discuss the supported browser versions. We can then use this browser list queries https://github.com/browserslist/browserslist. I guess that the build configuration will have to be separated for each target. However, I think that we can remove the import statements and we would not need the babel for NodeJs at all.

src/datasets.js
```diff
@@ -360,9 +359,10 @@ export default class Datasets extends Endpoint {
             method: 'GET',
             qs: query,
             json: false,
-            gzip: true,
             resolveWithFullResponse: true,
             encoding: null,
+            expBackoffMillis: 0,
+            expBackoffMaxRepeats: 0, // Turn off retries. We need to retry here because of bailing parse errors. //TODO: REFACTOR
```
mnmkng: What?

src/endpoint.js
```diff
@@ -1,7 +1,8 @@
 export default class Endpoint {
-    constructor(httpClient, basePath) {
+    constructor(httpClient, basePath, requestOptions) {
```
mnmkng: Hmm, thinking about it. The class should not be `Endpoint` but `EndpointGroup` or `Resource` because endpoints are the individual functions.

Then the options would be called `endpointGroupOptions` or `resourceOptions`.

I think I like `Resource` the best.

petrpatek:
I have just noticed in the API that there some new parameters for example `my` in https://docs.apify.com/api/v2#/reference/actors/actor-collection/get-list-of-actors.

petrpatek:
Meanwhile, I get more feedback I have rewritten the actor endpoint documentation. There are multiple ways shown how we could possibly structure the documentation. I would not include an example of every endpoint method. I think that some basic explanation with a link to the API docs and some basic param description is good enough. @jancurn

jancurn:
@petrpatek Definitely, for each function we just need basic description, link to API reference, and a list of parameters including types. That's all really. Examples should be in the main README only.

src/actors.js
```diff
@@ -98,8 +98,44 @@ export default class Actors extends Resource {
...
+     * //TODO: We could possible add only link to the endpoint
+     * ***Example**:
      *
+     * ```javascript
+     * const actor = {
+     *    "name": "MyActor",
+     *    "description": "My favourite actor!",
+     *    "isPublic": false,
+     *    "title": "My Actor",
+     *    "restartOnError": false,
```
mnmkng: I would definitely not add this, because it would get outdated sooner or later.

src/actors.js
```diff
@@ -247,10 +291,12 @@ export default class Actors extends Resource {
      *
      * The endpoint supports pagination using limit and offset parameters and it will not return more than 1000 array elements.
      *
+     * For more details see (get list of runs endpoint)[https://docs.apify.com/api/v2#/reference/actors/run-collection/get-list-of-runs]
+
      * @memberof ApifyClient.actors
      * @instance
      * @param {Object} options
-     * @param {String} options.actorId - Unique actor ID
+     * @param {String} options.actorId - Actor ID or a tilde-separated owner's username and actor name.
```
mnmkng: Actually `/` separated. That's why the `replaceSlashWithTilde` function is there. Ex.: `apify/web-scraper`

mnmkng:
@mtrunkat @jancurn I propose we test this thoroughly and merge it. There are things that need to be refactored for sure, and I have the plan to look at some of those myself, but I don't think running another review passes makes sense, especially given the fact that a lot of the code will be changed anyway. What do you think?

jancurn:
I agree, let's get it out. Do you want to release it as `latest` or `beta` ? I'm thinking we should let people know it's still in beta...

mtrunkat:
One thing - this new version of client doesn't have `client.setOptions()` which is used in couple of examples as https://help.apify.com/en/articles/1474989-how-to-get-put-records-into-a-named-key-value-store-in-an-actor . We will need to search for them and update before release.

src/http-client.js
```diff
@@ -114,6 +114,9 @@ export class HttpClient {
 
         if (options.headers && options.headers['content-type'] === null) {
             delete config.headers['content-type'];
+            delete this.axiosInstance.defaults.headers.post['Content-Type'];
+            delete this.axiosInstance.defaults.headers.put['Content-Type'];
+            delete this.axiosInstance.defaults.headers.patch['Content-Type'];
```
mnmkng: What if we create our own instance instead of using the default one? Would it still have those defaults?

mtrunkat:
One more thing - we will need option to disable retries on client as user can't wait 2 minutes to receive the error. So I would add some hard limit to options.

mnmkng:
@mtrunkat 

> One more thing - we will need option to disable retries on client as user can't wait 2 minutes to receive the error. So I would add some hard limit to options.

You can still set `expBackoffMaxRepeats` in the client constructor options. That should be enough, no?

mtrunkat:
@mnmkng that's exactly what I need, I missed this

mnmkng:
@mtrunkat Yeah, it's not exactly obvious. One of the reasons I want to do a refactoring pass on the client in April.

mtrunkat:
One more thing. If you use this in-browser then following 2 errors are logged to console:

```
Refused to set unsafe header "user-agent"
Refused to set unsafe header "accept-encoding"
```

We should not be passing these on the client-side.

mnmkng:
Close in favor of #146

PR #144, opened 2020-10-06
Refactor/v1

jbartadev:
- Added `call` method for actor and task clients
- Added `tests` for call method on actor and task both
- Added private `_call` method to resource client


src/resource_clients/actor.js
```diff
@@ -66,36 +66,23 @@ class ActorClient extends ResourceClient {
     async call(options = {}) {
         ow(options, ow.object.exactShape({
             contentType: ow.optional.string,
-            memoryMbytes: ow.optional.number,
-            timeoutSecs: ow.optional.number,
+            memory: ow.optional.number,
+            timeout: ow.optional.number,
```
mnmkng, line 70: Let's keep the `Mbytes` and `Secs`. It's in line with our naming. We always add the units to number variables.

src/resource_clients/actor.js
```diff
@@ -66,36 +66,23 @@ class ActorClient extends ResourceClient {
...
-            contentType,
-            waitForFinish: waitSecs,
-            memory: memoryMbytes,
-            build,
-            webhooks,
-            input,
-        };
+        const { timeout, ...callOptions } = options;
+
+        if (timeout >= 0) callOptions.timeout = timeout;
```
mnmkng, line 79: I'm thinking, instead of this, we could validate that the `timeoutSecs` directly in `ow` to be non-negative.

mnmkng: Same for `waitSecs` actually.

src/resource_clients/task.js
```diff
@@ -53,34 +53,22 @@ class TaskClient extends ResourceClient {
 
     async call(options = {}) {
         ow(options, ow.object.exactShape({
-            memoryMbytes: ow.optional.number,
-            timeoutSecs: ow.optional.number,
+            memory: ow.optional.number,
+            timeout: ow.optional.number,
```
mnmkng, line 57: same here

test/tasks.test.js
```diff
@@ -189,6 +189,56 @@ describe('Task methods', () => {
...
+            const body = { data };
+            const waitSecs = 1;
+
+            const query = {
+                timeout,
+                memory,
+                build,
+            };
+
+            mockServer.setResponse({ body }); // this is not used for the task.start()
```
mnmkng, line 210: actually, here it is used for `task.start()` too

mnmkng (commented):
Looks good. Some minor improvements and it's good to go. Thanks a lot @jbartadev, it was pretty painful.

src/resource_clients/actor.js
jbartadev: Ok, make sense. Renamed.

src/resource_clients/actor.js
jbartadev: Added the non-negative validation, removed the condition 👍

src/resource_clients/task.js
jbartadev: sure.

mnmkng (approved):

PR #150, opened 2020-12-30
Add anchors to headings in template

m-murasovs:
Really don't know if it's gonna be worth the effort, but I've cobbled together this anchor-adding template.

The only thing (I think) I can't figure out is in the comment

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ All public classes, methods and their parameters can be inspected in this API re
 
 <a name="ApifyClient"></a>
 
-### [](#ApifyClient) ApifyClient
+### [](#apifyclient) ApifyClient
 
 ApifyClient is the official library to access [Apify API](https://docs.apify.com/api/v2) from your
 JavaScript applications. It runs both in Node.js and browser.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -163,9 +164,10 @@ JavaScript applications. It runs both in Node.js and browser.
 
 * * *
 
-<a name="new_ApifyClient_new"></a>
+<a name="ApifyClient"></a>
```
m-murasovs, line 167: I really can't figure out how to do this one, since 'new' isn't included in the props

Screenshot 2020-12-30 at 19 14 01

README.md
```diff
@@ -163,9 +164,10 @@ JavaScript applications. It runs both in Node.js and browser.
...
+
+#### [](#ApifyClient) ApifyClient
 
-#### `new ApifyClient([options])`
```
m-murasovs, line 171: Oh, and here

mnmkng:
It seems that the original `header.hbs` gets the info in the `anchorName` helper.

```hbs
<a name="{{{anchorName}}}"></a>
```

Which can be found in `node_modules/dmd/helpers/ddata.js` and it extracts the information from a `this.kind` property:

image

So I guess by looking at the `this.kind === 'constructor'` you could work out the last two.

PR #159, opened 2021-02-15
SetRecord: Removed uploading using signed URL

drobnikj:


src/utils.js
```diff
@@ -70,11 +70,13 @@ function stringifyWebhooksToBase64(webhooks) {
...
- * @param {*} value
+ * Gzip provided value, otherwise returns null.
+ * @param value
+ * @return {Promise<null|Buffer>}
  */
 async function maybeGzipValue(value) {
-    if (!isNode()) return value;
-    if (value == null) return value;
+    if (!isNode()) return null;
+    if (value == null) return null;
```
mnmkng, line 79: Please use `return` or `return undefined`. I don't think we return `null` anywhere in the client.

mnmkng (changes requested):
Looks good. Just a tiny nitpick.

mtrunkat (approved):

mnmkng (approved):

PR #163, opened 2021-03-03
Feature/resurrect run with options

nguyeda1:
Resurrect run with options:
- build (tag or version)
- timeout (in seconds)
- memory (in Mbytes)
- resendRunCreatedWebhook (boolean if resurrect should resend Run Created webhook)

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,13 +171,13 @@ JavaScript applications. It runs both in Node.js and browser.
 
 | Param | Type | Default |
 | --- | --- | --- |
-| [options] | <code>object</code> |  |
-| [options.baseUrl] | <code>string</code> | <code>&quot;https://api.apify.com&quot;</code> |
-| [options.maxRetries] | <code>number</code> | <code>8</code> |
-| [options.minDelayBetweenRetriesMillis] | <code>number</code> | <code>500</code> |
-| [options.requestInterceptors] | <code>Array.&lt;function()&gt;</code> |  |
-| [options.timeoutSecs] | <code>number</code> |  |
-| [options.token] | <code>string</code> |  |
+| [options] | <code>object</code> |  | 
+| [options.baseUrl] | <code>string</code> | <code>&quot;https://api.apify.com&quot;</code> | 
+| [options.maxRetries] | <code>number</code> | <code>8</code> | 
+| [options.minDelayBetweenRetriesMillis] | <code>number</code> | <code>500</code> | 
+| [options.requestInterceptors] | <code>Array.&lt;function()&gt;</code> |  | 
+| [options.timeoutSecs] | <code>number</code> |  | 
+| [options.token] | <code>string</code> |  | 
 
 
 * * *
@@ -200,7 +200,7 @@ https://docs.apify.com/api/v2#/reference/actors/actor-object
 
 | Param | Type |
 | --- | --- |
-| id | <code>string</code> |
+| id | <code>string</code> | 
 
 
 * * *
@@ -214,7 +214,7 @@ https://docs.apify.com/api/v2#/reference/actor-builds/build-object
 
 | Param | Type |
 | --- | --- |
-| id | <code>string</code> |
+| id | <code>string</code> | 
 
 
 * * *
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -171,13 +171,13 @@ JavaScript applications. It runs both in Node.js and browser.
...
-| [options.requestInterceptors] | <code>Array.&lt;function()&gt;</code> |  | 
-| [options.timeoutSecs] | <code>number</code> |  | 
-| [options.token] | <code>string</code> |  | 
+| [options] | <code>object</code> |  |
+| [options.baseUrl] | <code>string</code> | <code>&quot;https://api.apify.com&quot;</code> |
+| [options.maxRetries] | <code>number</code> | <code>8</code> |
+| [options.minDelayBetweenRetriesMillis] | <code>number</code> | <code>500</code> |
+| [options.requestInterceptors] | <code>Array.&lt;function()&gt;</code> |  |
+| [options.timeoutSecs] | <code>number</code> |  |
+| [options.token] | <code>string</code> |  |
```
mnmkng, line 180: What's up with all the whitespace?

src/resource_clients/run.js
```diff
@@ -92,13 +92,26 @@ class RunClient extends ResourceClient {
 
     /**
      * https://docs.apify.com/api/v2#/reference/actor-runs/resurrect-run/resurrect-run
+     * @param {object} [options]
+     * @param {object} [options.contentType]
+     * @param {string} [options.build]
+     * @param {number} [options.memory]
+     * @param {number} [options.timeout]
+     * @param {boolean} [options.resendRunCreatedWebhook]
```
mnmkng, line 100: I don't know if this is a good idea to have a property for one specific webhook event. This could get out of hand pretty fast. But that's a question of the API design, not really the client.

mnmkng: I mean, it looks like the property is a hotfix of something. Which it technically is.

mtrunkat: We have 2 options - we will do a breaking change and start sending the "RUN STARTED" event webhook on each resurrect or we make this opt-in in a non-breaking way

README.md
nguyeda1: I think my eslint config automatically removes unnecessary whitespace.. is it meant to be there?

mnmkng: I guess it does not matter functionally, but I when someone generates it next time, it'll add the whitespace again and we'll have all the changes visible  in the PR again, so I'd prefer if it stayed the same.

src/resource_clients/run.js
mnmkng: Yeah, agree making it opt-in. But can we make it `resendWebhooks: true` or `resendWebhooks: ['RUN.CREATED' , 'RUN.FAILED']? This option is a bit weird.

mtrunkat: The thing is that all other except created are being resent always, just created one not

mnmkng: Yeah I remember, so maybe the question is if that behavior is correct and if we should allow the user to tweak it.

I mean, it's not critical at all, but it's definitely super weird to have a super specific option like that 😄 If you're OK with it in the API, I'm ok with it in the client.

README.md
nguyeda1: @mnmkng I see.. I generated the Readme via the npm script.. should be fixed now

nguyeda1:
@mtrunkat @mnmkng removed resendRunCreatedWebhook property based on [this conversation](https://apifier.slack.com/archives/CA6EBU3CM/p1614934864035000)

mnmkng (approved):

mnmkng:
Oh, could you please also update the version and changelog @nguyeda1 ? This is a new feature so the next version should be `1.1.0`, not `1.0.6` Thanks

nguyeda1:
> Oh, could you please also update the version and changelog @nguyeda1 ? This is a new feature so the next version should be `1.1.0`, not `1.0.6` Thanks

@mnmkng done :)

PR #174, opened 2021-06-01
chore: update apify-shared

B4nan:
Updates `apify-shared` to use only the needed packages (`log` and `consts`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+1.2.4 / 2021/06/01
+===================
+- use new `apify-shared` packages to reduce bundle size
+
 1.2.3 / 2021/05/27
 ===================
 - Fixed invalid max body length setting thanks to a transitive default in `axios`.
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,6 @@
 const ow = require('ow').default;
-const { ME_USER_NAME_PLACEHOLDER } = require('apify-shared/consts');
-const logger = require('apify-shared/log');
+const { ME_USER_NAME_PLACEHOLDER } = require('@apify/consts');
+const { default: logger } = require('@apify/log');
 
 const HttpClient = require('./http_client');
 const Statistics = require('./statistics');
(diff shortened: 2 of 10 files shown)


src/index.js
```diff
@@ -1,6 +1,6 @@
 const ow = require('ow').default;
-const { ME_USER_NAME_PLACEHOLDER } = require('apify-shared/consts');
-const logger = require('apify-shared/log');
+const { ME_USER_NAME_PLACEHOLDER } = require('@apify/consts');
+const { default: logger } = require('@apify/log');
```
mnmkng, line 3: This is cool. It implies that it's the default logger instance 😄

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+1.2.4 / 2021/06/01
+===================
+- use new `apify-shared` packages to reduce bundle size
```
mnmkng, line 3: Have you checked how large the reduction was?

src/resource_clients/key_value_store.js
```diff
@@ -1,5 +1,5 @@
 const ow = require('ow').default;
-const log = require('apify-shared/log');
+const log = require('@apify/log');
```
mnmkng, line 2: In `index.js` it's imported from `default` and here it's imported directly 🤷

B4nan: good catch, apparently this is not tested

mnmkng: Yeah, I don't think we're very thorough in testing loggers and logging 😞

CHANGELOG.md
B4nan: hmmm currently it actually did not make any difference, because the original `apify-shared` is still transitive dependency of some packages like `browser-pool` or the SDK. if i remove it, we save ~1mb (from ~207mb), but the main benefit will be in removed dependencies, so i am expecting much more (lets say tens of mbs).

mnmkng: I was talking about the `apify-client` webpack bundle size 😄

mnmkng (approved):

PR #176, opened 2021-06-15
Feature/run storages

nguyeda1:
Run storages are available via `/v2/actor-runs/{runId}/{key-value-store | log | request-queue | dataset}` .. see https://docs.apify.com/api/v2#/reference/actor-runs/run-object-and-its-storages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+1.2.5 / 2021/06/15
+===================
+- Enabled access to actor run storages via `RunClient`
+
 1.2.4 / 2021/06/01
 ===================
 - use new `apify-shared` packages to reduce bundle size
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "apify-client",
-    "version": "1.2.4",
+    "version": "1.2.5",
     "description": "Apify API client for JavaScript",
     "main": "src/index.js",
     "keywords": [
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+1.2.5 / 2021/06/15
+===================
+- Enabled access to actor run storages via `RunClient`
```
mnmkng, line 3: Should be a minor version bump to `1.3.0`, since features were added.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "apify-client",
-    "version": "1.2.4",
+    "version": "1.2.5",
```
mnmkng, line 3: dtto

mnmkng (approved):
Ok, now I'm a bit torn. I thought you would need to add some code, but I scrolled down and it seems that you only added docs and the code was already there. So I guess you can ignore my comments about the version 1.3.0... 🤷‍♂️

drobnikj (changes requested):
Please add simple tests.
There is a skeleton prepared already.
https://github.com/apify/apify-client-js/blob/master/test/runs.test.js#L204

nguyeda1:
Added the same tests as for `actor().lastRun()`

nguyeda1:
> Please add simple tests.
> There is a skeleton prepared already.
> https://github.com/apify/apify-client-js/blob/master/test/runs.test.js#L204

Added the same tests as for actor().lastRun()

drobnikj (approved):

PR #184, opened 2021-07-29
feat: changed passing token using request HTTP header instead of the request parameter for every API calls

drobnikj:
- [ ] Run integrations test with the new version of the client

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - master
+      # This will allow releasing beta versions from non-master releases.
+      # TODO: We need to release this version with other tags than beta. It can mess up the beta version as it will override it with older code.
       - version/**
   # A release via GitHub releases will deploy a latest version
   release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-1.4.0 / 2021/07/29
+1.4.0 / 2021/08/24
 ===================
 - Changed passing token using request HTTP header instead of the request parameter for every API calls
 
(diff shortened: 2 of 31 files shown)


drobnikj:
I just realize that I can probably remove all `this._params()` calls. We probably do not need it in the future.

src/base/resource_client.js
```diff
@@ -30,7 +30,7 @@ class ResourceClient extends ApiClient {
         const requestOpts = {
             url: this._url(),
             method: 'GET',
-            params: this._params(options),
+            params: options,
```
mnmkng, line 33: We should make a copy of the options where we're passing the options object directly and not recreating or destructuring. This is asking for weird bugs in user code.

mnmkng (changes requested):
Looks good besides that one nitpick 👍

mtrunkat (approved):

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-1.4.0 / 2021/07/29
+1.4.0 / 2021/08/24
```
B4nan, line 1: next release will be 2.0, this needs rebase

drobnikj: We decided to release it as 1.4.0 before the TS rewrite

B4nan: master is already v2, so you will need to create a new branch from latest 1.3 and target that instead

mnmkng (approved):

B4nan (approved):

B4nan:
cc @vladfrangu this will need to be backported to master once merged

.github/workflows/release.yml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - master
+      # This will allow releasing beta versions from non-master releases.
+      # TODO: We need to release this version with other tags than beta. It can mess up the beta version as it will override it with older code.
```
B4nan, line 9: mmm good catch, but you should have addressed this before merging, as beta releases are automated based on commits to those branches

PR #185, opened 2021-08-12
chore: get TS base up and running, update deps

vladfrangu:


src/http_client.js
```diff
@@ -11,7 +11,9 @@ const {
 const {
     isNode,
 } = require('./utils');
-const { version } = require('../package.json');
+
+// eslint-disable-next-line no-eval
+const { version } = eval(`require('../package.json')`);
```
szmarczak, line 16: Webpack... I guess??

vladfrangu: tsc actually, since otherwise it tends to clone the package.json. It's ugly, I know, and I'll see tomorrow what I can do to solve the need for this (since it might also make webpack angry)

szmarczak: Okay. Not sure if this is still relevant but https://twitter.com/kamilogorek/status/1102272038411137025

vladfrangu: That's very insightful, thanks! It's actually gonna be weird..In webpack I need to let it do it's thing, in TS i need it to just require as is, without importing... I'm sure I'll figure something out 👀

vladfrangu: Was quite a headache, but I figured out how to make both worlds happy 🎉

webpack.config.js
```diff
@@ -17,16 +20,16 @@ module.exports = {
     resolve: {
         mainFields: ['browser', 'main', 'module'],
         extensions: ['*', '.js'],
+        fallback: {
+            fs: false,
+            os: false,
+            stream: false,
+            util: false,
+            zlib: false,
+        },
```
mnmkng, line 29: Not sure how webpack 5 deals with Node internals but please check the bundle size between the versions. E.g. `Buffer: false` was there earlier to remove a pretty large `Buffer` polyfill.

webpack.config.js
```diff
@@ -1,5 +1,8 @@
-const BrotliPlugin = require('brotli-webpack-plugin');
```
mnmkng, line 1: Webpack 5 supports brotli out of the box? Or why remove it?

vladfrangu: It wasn't even used in the config, it was commented out 👀

webpack.config.js
vladfrangu: Almost all nodejs APIs have been removed from webpack 5, and you need to manually add them (like I had to do with process). 

Nonetheless, here's an old vs new comparison (we actually are saving some bytes)
image

mnmkng: Hmm, that looks big. We must have been doing something wrong with this bundle. https://bundlephobia.com/package/apify-client@1.3.0 It's used only for testing so maybe it's not that important, but still.

webpack.config.js
mnmkng: Oh, I see. That explains the bundle size. Yeah, let's keep it removed. At least we won't have to inflate it before using in tests.

webpack.config.js
vladfrangu: It is now almost 4x smaller (almost..)
image

.eslintrc.json
```diff
@@ -1,3 +1,26 @@
...
+    "project": "tsconfig.eslint.json"
+  },
+  "overrides": [
+    {
+      "files": [
+        "src/**/*.ts"
+      ],
+      "rules": {
+        "import/extensions": 0
+      }
```
mnmkng, line 15: is this still needed?

package.json
```diff
@@ -1,92 +1,106 @@
...
+	"module": "dist/index.mjs",
+	"types": "dist/index.d.ts",
+	"browser": "dist/bundle.js",
+	"unpkg": "dist/bundle.js",
+	"exports": {
+		".": {
+			"import": "./dist/index.mjs",
+			"require": "./dist/index.js"
+		}
+	},
```
mnmkng, line 15: I think this was causing issues for us in `apify-shared`. The Apify Web could not be built because webpack was not finding the right files. I think we fixed it, but another user on Discord complained that he can't use the client with a `create-react-app`, where he does not have access to Webpack config...

vladfrangu: I've added entries to the package.json that should point browser compilers to use our bundle... Now, I cannot test it sadly, but it shouldn't cause issues..

.eslintrc.json
vladfrangu: Brought out the 🧹 and removed it

package.json
```diff
@@ -1,92 +1,106 @@
...
-        "morgan": "^1.10.0",
-        "puppeteer": "^5.5.0",
-        "webpack": "^4.44.2",
-        "webpack-cli": "^3.3.12"
-    }
+	"name": "apify-client",
+	"version": "1.3.0",
+	"description": "Apify API client for JavaScript",
+	"main": "dist/index.js",
+	"module": "dist/index.mjs",
```
B4nan, line 6: i'd be careful with the ESM support. i had to remove it from the apify-shared packages as it caused issues when used in downstream with webpack (it was easier to just remove it, not gonna lie, i am a webpack noobie, so i might just have missed something)

it might as well be caused by webpack 4 but as we are talking about our internal packages, we cant just force others in the company to upgrade webpack to v5 just to be able to use new packages that we ourselves create.

i'll try to look up what was the exact issue

vladfrangu: Could you check if https://github.com/apify/apify-client-js/pull/185#discussion_r688477222 is valid? It's what we do in other personal projects and it should just work..
Also hrm..yeah that would be a problem, upgrading to webpack 5 was easy here, I don't know how much it'll affect other company projects. Needs testing 😓

B4nan: ok now i remember - the thing was that guys are using cypress and it was not possible to override the webpack config it uses.

it was not even possible to read the consts, here is the failed build and one of the errors:

> Cannot read property 'ACTOR_BASE_DOCKER_IMAGES' of undefined

https://github.com/apify/apify-web/pull/1169/checks?check_run_id=2847671842

the workaround i came accross was `mainFields: ['browser', 'main', 'module'],` in the webpack config, which basically prioritizes the CJS version instead of ESM one - but apparently it was not possible to do so via cypress. 

(we can definitely keep it here for now, and remove it as a last resort if we encounter similar issues in downstream)

B4nan: > Could you check if #185 (comment) is valid? It's what we do in other personal projects and it should just work..

yes, @mnmkng is talking about the very same thing. but we did not fix it, the "fix" was to remove the ESM support from apify-shared.

https://github.com/apify/apify-shared-js/pull/199

i wouldn't be surprised if its a webpack 4 issue, also havent tested that myself extensively

vladfrangu: I mean worst case, I can remove the esm parts for now, as I can't test out the new version locally to see if it works or not. Thoughts?

mnmkng: Please update changelog and version and we can merge it. It will release beta and we can test it.

vladfrangu: Done 👍

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+2.0.0-alpha0 / TBD
```
mnmkng, line 1: no need to use the `-alpha` in changelog. The CI will automatically add beta tags to releases.

mnmkng (approved):

PR #186, opened 2021-08-13
fix: build module before publishing in CI

vladfrangu:


B4nan (approved):
btw that's what the `prepare` script was for, as it trigger the build after install

vladfrangu:
> btw that's what the `prepare` script was for, as it trigger the build after install

Yeah.. some repositories moved to prepublishOnly (as to not slow down installs), but I see in this repo the prepublishOnly script is a bash one (and I'll need to sharpen my bash knowledge/write a quick js script for crossplatform compat to solve it). Let me know what's ideal 👍

B4nan:
i am fine with how it is now, the build should be done explicitly in CI as it is and the `prepublishOnly` do not have to be crossplatform ready, as publishing is only meant to be executed in CI where we control the environment

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "apify-client",
-	"version": "2.0.0",
```
mnmkng, line 3: No need to bump version. We have not released 2.0 yet.

mnmkng (approved):

PR #188, opened 2021-08-16
refactor: convert api base clients to TS

vladfrangu:
Depends on #187


src/base/resource_client.ts
```diff
@@ -16,17 +16,10 @@ const MAX_WAIT_FOR_FINISH = 999999;
...
  */
-class ResourceClient extends ApiClient {
-    /**
-     * @param {object} [options]
-     * @return {Promise<?object>}
-     * @private
-     */
-    async _get(options = {}) {
+export class ResourceClient extends ApiClient {
+    protected async _get<T extends Record<string, unknown>, R extends unknown>(options: T = {} as T): Promise<R | undefined> {
```
B4nan, line 22: does it actually make any difference to define `R extends unknown`? `unknown` is a top type, every type is its extension, so it feels redundant

src/base/resource_client.ts
```diff
@@ -117,12 +99,15 @@ class ResourceClient extends ApiClient {
         } while ((shouldRepeat()));
```
B4nan, line 99: i know its unrelated, but :]
```suggestion
        } while (shouldRepeat());
```

B4nan (approved):
looking good, i am just not sure about the `extends unknown`, won't hurt anyone, but i don't think it has any value

src/base/resource_client.ts
vladfrangu: oh yeah, good point. Can simplify it to just `R` 👍

B4nan (approved):

PR #189, opened 2021-08-16
refactor: convert utils to TS

vladfrangu:


src/utils.ts
```diff
@@ -67,54 +83,40 @@ function stringifyWebhooksToBase64(webhooks) {
...
-async function maybeGzipValue(value) {
+export async function maybeGzipValue(value: unknown): Promise<Buffer | undefined> {
     if (!isNode()) return;
     if (value == null) return;
 
     // Request compression is not that important so let's
     // skip it instead of throwing for unsupported types.
     const areDataStringOrBuffer = (typeof value === 'string') || Buffer.isBuffer(value);
-    const areDataLargeEnough = Buffer.byteLength(value) >= MIN_GZIP_BYTES;
+    const areDataLargeEnough = Buffer.byteLength(value as string) >= MIN_GZIP_BYTES;
```
szmarczak, line 99: Shouldn't this be

```ts
if (!isNode()) return;
if (typeof value !== 'string' || !Buffer.isBuffer(value)) return;

const areDataLargeEnough = Buffer.byteLength(value) >= MIN_GZIP_BYTES;
if (areDataLargeEnough) {
	return gzipPromise(value);
}
```

This way it won't throw on `maybeGzipValue({})`

szmarczak (approved):

src/utils.ts
vladfrangu: You're right, that would be better

szmarczak: The `if (value == null) return;` check seems to be unnecessary now

szmarczak (approved):

B4nan (approved):

PR #190, opened 2021-08-16
revert: replace just the first `/` with a `~` in ids

vladfrangu:


src/base/api_client.ts
```diff
@@ -70,7 +70,7 @@ export abstract class ApiClient {
     }
 
     protected _toSafeId(id: string): string {
-        return id.replace(/\//g, '~');
+        return id.replace('/', '~');
```
szmarczak, line 73: Can you please add a comment why's that? So we don't go into the same pitfall again :)

vladfrangu: Would a link to the slack thread be enough? Should I explain it instead or?

szmarczak: A comment would be best. Something like `The id looks like 'username/project' so we need to replace the slash only once` or so.

vladfrangu: Actually, probably explaining it would be the best so other contributors can see it at a glance

B4nan (approved):

szmarczak (approved):

PR #192, opened 2021-08-18
refactor: convert Builds to TS

vladfrangu:
Depends on #191


src/resource_clients/actor.ts
```diff
@@ -1,72 +1,62 @@
...
-    /**
-     * @param {ApiClientOptions} options
-     */
-    constructor(options) {
+export class ActorClient extends ResourceClient {
+    constructor(options: ApiClientOptions) {
         super({
-            resourcePath: 'acts',
             ...options,
+            resourcePath: 'acts',
```
szmarczak, line 28: Why the change?

src/resource_clients/actor_collection.ts
```diff
@@ -0,0 +1,65 @@
...
+import { ActorVersion } from './actor_version';
+
+/**
+ * @hideconstructor
+ */
+export class ActorCollectionClient extends ResourceCollectionClient {
+    constructor(options: ApiClientOptions) {
+        super({
+            ...options,
+            resourcePath: 'acts',
```
szmarczak, line 15: same here

szmarczak (approved):

src/resource_clients/actor.ts
vladfrangu: Wrong PR to leave a review about this, but because of this.. I will need to check if the type maybe is wrong and I should use the subresource one instead
image

vladfrangu: And the answer is yes, I should be using the other type.. I'll fix in that PR, and rebase appropriately

B4nan: already merged the other one, so just fix it in this one pls

B4nan (approved):

PR #193, opened 2021-08-18
refactor: convert Dataset to TS

vladfrangu:


src/resource_clients/dataset.ts
```diff
@@ -156,29 +123,75 @@ class DatasetClient extends ResourceClient {
...
+    name: string;
+    userId: string;
+    createdAt: string;
+    modifiedAt: string;
+    accessedAt: string;
+    itemCount: number;
+    cleanItemCount: number;
+    // TODO: Are these actually always just empty objects?
+    actId?: Record<string, unknown> | null;
+    actRunId?: Record<string, unknown> | null;
```
B4nan, line 152: i thought those are strings, not objects. at least in SDK tests they are used as strings.

src/resource_clients/dataset.ts
```diff
@@ -156,29 +123,75 @@ class DatasetClient extends ResourceClient {
...
+    fields?: string[];
+    // TODO: the api seems to return a stats field, that isn't documented
```
B4nan, line 154: cc @mtrunkat, can you please shed some light on this, or suggest someone who can help?

(basically all the questions marked with TODO in this PR)

src/resource_clients/dataset.ts
vladfrangu: The api tester on the docs site returns an empty object, the api examples say they're null, so.. some clarification would be lovely!

src/resource_clients/dataset.ts
```diff
@@ -156,29 +123,75 @@ class DatasetClient extends ResourceClient {
...
- * @typedef {object} PaginationList
- * @property {object[]} items - List of returned objects
- * @property {number} total - Total number of objects
- * @property {number} offset - Number of objects that were skipped
- * @property {number} count - Number of returned objects
- * @property {number} limit - Requested limit
- */
+export interface Dataset {
+    id: string;
+    name: string;
```
metalwarrior665, line 143: name is definitely optional

src/resource_clients/dataset.ts
```diff
@@ -156,29 +123,75 @@ class DatasetClient extends ResourceClient {
...
+    userId: string;
+    createdAt: string;
+    modifiedAt: string;
+    accessedAt: string;
+    itemCount: number;
+    cleanItemCount: number;
```
metalwarrior665, line 149: Some old datasets (like 2 years ago) don't have this field, not sure if we want to just forget about it?

src/resource_clients/dataset.ts
metalwarrior665: Haha, I think the API tester should get a big rewamp:)

src/resource_clients/dataset.ts
vladfrangu: Couldn't a migration be written for those? Otherwise, I can mark the field as possibly undefined but then the docs also need an update.

B4nan: i guess we should model it based on how it works now, 2 years is quite a long time, and if should not matter much anyway

src/resource_clients/dataset.ts
metalwarrior665: What docs are you using exactly? https://docs.apify.com/api/v2 have a reputation of being terribly outdated even though I pointed that out several times.

metalwarrior665: E.g. https://api.apify.com/v2/datasets/8kUL85wh7Zhu9KkJx

vladfrangu: > What docs are you using exactly? [docs.apify.com/api/v2](https://docs.apify.com/api/v2) have a reputation of being terribly outdated even though I pointed that out several times.

That's the docs I'm using. 😱
Do you have any that might be closer-to-current-returns, since throwing random API calls isn't the best idea / ideal..

metalwarrior665: Haha, this is what I got months ago when testing :) https://github.com/metalwarrior665/apify-client-rs/blob/master/src/datasets.rs#L139
But since we probably don't have a real source of truth, it will be a combination of:
- Finding some types in the SDK
- Checking the API source code
- Testing API calls
- Asking around
- Forcing platform team to finally type the API :)

src/resource_clients/dataset.ts
metalwarrior665: I agree to let it be always a number. I just run into it when testing some old saved dataset but only a tiny tiny fraction of users have these saved somewhere and actively using them.

src/resource_clients/dataset.ts
vladfrangu: > Forcing platform team to finally type the API :)

🥺 

Yeah, I'll be.. marking it as possibly undefined for now, and after I go through them all run a pass with the API, see what it returns, even if that'll be pain. Maybe I can read the API source instead and (🤞) gain a better idea..

metalwarrior665: I mean the `name` case is simple. It is a string for named datasets and undefined for unnamed

vladfrangu: Yep, that one I fixed in [`103c03d` (#193)](https://github.com/apify/apify-client-js/pull/193/commits/103c03d601543d70d3d0b1e52036f07f50a2b016), my remark was more general (since it seems there's even discrepancy between the api docs and tester) 😄

B4nan (approved):

B4nan:
some tests failing on this one

vladfrangu:
Yep, CI is failing while we wait on https://github.com/apify/apify-core/issues/3503 to.. progress

vladfrangu:
To not prevent the rewrite to continue, I've added a temporary default to whatever the user provided. Should be fine now 👍

PR #195, opened 2021-08-18
refactor: convert KeyValue to TS

vladfrangu:


src/resource_clients/key_value_store.ts
```diff
@@ -180,11 +177,48 @@ class KeyValueStoreClient extends ResourceClient {
     }
 }
 
-module.exports = KeyValueStoreClient;
+export interface KeyValueStore {
+    id: string;
+    name: string;
```
B4nan, line 182: this is the same as discussed in the other PR. names are optional (no name = default dataset/kv store/...)

same with the `actId` and `actRunId`, those are definitely strings, not objects

B4nan (approved):

PR #200, opened 2021-08-23
refactor: convert schedule to TS

vladfrangu:


src/resource_clients/actor.ts
```diff
@@ -54,9 +54,8 @@ export class ActorClient extends ResourceClient {
     /**
      * Starts an actor and immediately returns the Run object.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
-     * @return {Promise<Run>}
      */
-    async start(input: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
+    async start(input?: string | Buffer, options: ActorStartOptions = {}): Promise<ActorRun> {
```
szmarczak, line 58: This can be a stream as well https://github.com/axios/axios/blob/199c8aab64c45532389f3e56a913017a54803dfe/lib/adapters/http.js#L70-L82

B4nan (approved):

src/resource_clients/actor.ts
vladfrangu: Hrm... Could add ReadableStream to the types I suppose. Thoughts @B4nan @mnmkng?

mnmkng: This is a bit hard because the package is isomorphic (node+browser). So I guess even the `Buffer` is problematic. Not sure how this is handled in TS.

szmarczak: I'd leave this as `unknown` then...

vladfrangu: Yeah... that's probably the better move, keeping it as `unknown`

B4nan: > This is hard because the package is isomorphic (node+browser).

does this really matter? we are talking about types, those are not used when you run the code.

not sure if this will fail without `@types/node` or if it would result in `any/unknown` instead tho

fwiw axios does not use `Buffer` in their typings and [use just `any` for `data`](https://github.com/axios/axios/blob/master/index.d.ts#L59)

PR #202, opened 2021-08-24
feat: Add clients for actor env vars

jirimoravcik:
This PR adds clients for the new env vars API.
I added TODOs to the docs links, because as of now, the docs are not created.

@mnmkng, What is the suggested procedure here - create docs first and then merge at the same time or can we just release new client and add the docs later?


mnmkng:
@jirimoravcik Looks like the tests are failing for some reason. Otherwise it looks good. We can definitely release the new client even before we add the API docs. The question is, Client v2 is not stable yet, so if you need this released fast, we should merge it and then backport it to the `version/1.x` branch and release from there.

jirimoravcik:
> Weird. I think they passed on the last commit to master. @vladfrangu can you please check this when you have a moment?

Yeah, but the weird thing is that this repository doesn't have a `package-lock.json` file, so that's the cause 99%. Is there a reason for that?

B4nan:
Finally a man of manners! Yes please, we should use lock files (together with lockfile maintenance from renovate) to actually find those issues early and have it under control (plus the benefit of caching node modules and making CI faster).

I can confirm head/master is failing the same way. Can't really tell whether we should adjust the assertions or if it's important for us to have the `;charset=utf-8` suffix in content type header.

vladfrangu:
It's an update to axios (or a downstream module in it) that's causing this issue. Pinning it to the version we're on right now in package.json "solves" it, but the better question is the one @B4nan just made

mnmkng:
Well if we had the deps locked, we would've not learned about the problem this fast, would we? And all the users would be installing a broken library. Thanks to not having the deps locked, we learned about the issue immediately. But let's leave this discussion for some other time. For now, just fix the problem.

vladfrangu:
So then what's the intended fix? Drop the suffix (is it even necessary), or pin axios to a version?

B4nan:
> Well if we had the deps locked, we would've not learned about the problem this fast, would we? 

That is where the lockfile maintenance steps in, we would know it as there would be a PR telling us the lockfile update failed.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+2.1.0 / TBD
+===================
+- Added `ActorEnvVarCollectionClient` and `ActorEnvVarClient`
+
```
B4nan, line 4: 2.0.0 is not out yet, so move this to that version

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "apify-client",
-	"version": "2.0.1",
+	"version": "2.1.0",
```
B4nan, line 3: no need for version bump as 2.0.0 is not yet out

B4nan (changes requested):
fixed the assertions in master and merged it here, so things should be passing

left few comments, not sure what to do with the docs?

package.json
mnmkng: 2.0.0 is out on NPM as far as I know.

vladfrangu: The beta is
image

mnmkng: It's not tagged as it was published by accident, but it's there 
image

jirimoravcik:
@B4nan @mnmkng @vladfrangu

jirimoravcik:
> i'll be happy to merge if we can make it pass the CI checks (and rebase to remove conflicts)
> 
> also I see some TODOs in there, without actual description/action steps - are those something you should resolved before we merge, or something for others? (who?)

conflicts & pipeline fixed. the todos are there because the documentation has not been created yet. I could just remove the todos?

B4nan:
let's keep them there i guess

PR #204, opened 2021-08-24
fix: revert back passing params using ApifyClient _params method

drobnikj:
I get back `this._params(options) because without it the internal sub-clients didn't get params from the parent client.

Sadly tests were not cover this case, so I updated the lastRun test to cover this case.

This reverts commit c766a50


B4nan (approved):
looking good but we should also link the reverted commit in the commit message

mnmkng (approved):

PR #210, opened 2021-08-26
refactor: finish conversion to TS

vladfrangu:


mnmkng:
> > > Should we also jump sail to nodejs 15+ here?
> > 
> > 
> > Would be nice. Wdyt?
> 
> I'm down for this anytime! I'm all in favor of supporting just the latest versions

We can bump to 14 but not higher because the Apify platform runs on Node 14.

vladfrangu:
> > > > Should we also jump sail to nodejs 15+ here?
> > > 
> > > 
> > > Would be nice. Wdyt?
> > 
> > I'm down for this anytime! I'm all in favor of supporting just the latest versions
> 
> We can bump to 14 but not higher because the Apify platform runs on Node 14.

So...should I make CI test on 14 and 16? Just to be sure both work?

mnmkng:
Yeah go ahead.

mnmkng (approved):

PR #211, opened 2021-09-03
feat: Implement method for request batch insert endpoint

valekjo:
This PR adds support for new batch request insert endpoint.

Two public methods are added:
 - ~~`batchAddRequests`, which just calls the endpoint and does not have any extra logic~~
 - `batchAddRequests`~~`Concurrently`~~, which accepts more requests than batch limit (25) and calls the endpoint multiple times (concurrently) and retries for unprocessed items.

The endpoint should not yet be listed anywhere.
* tests will be added to integration tests of app and will replace the current ones that call the endpoint directly

**We want to add this in experimental mode, so we can start using it and see if this is the way to go.**

### DIFF
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -156,7 +156,10 @@ export class RequestQueueClient extends ResourceClient {
 
     /**
      * Writes multiple requests to request queue concurrently in batches.
-     * THIS METHOD IS EXPERIMENTAL AND NOT YET INTENDED FOR USE.
+     * THIS METHOD IS EXPERIMENTAL AND NOT INTENDED FOR PRODUCTION USE.
+     *
+     * @private
+     * @experimental
      */
     async batchAddRequests(
         requests: Omit<RequestQueueClientRequestSchema, 'id'>[],
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -97,31 +97,6 @@ export class RequestQueueClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
-    async batchAddRequests(
-        requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
-        options: RequestQueueClientAddRequestOptions = {},
-    ): Promise<RequestQueueClientBatchAddRequestsResult> {
-        ow(requests, ow.array.ofType(ow.object.partialShape({
-            id: ow.undefined,
-        })).minLength(1).maxLength(25));
-
-        ow(options, ow.object.exactShape({
-            forefront: ow.optional.boolean,
-        }));
-
-        const response = await this.httpClient.call({
-            url: this._url('requests/batch'),
-            method: 'POST',
-            data: requests,
-            params: this._params({
-                forefront: options.forefront,
-                clientKey: this.clientKey,
-            }),
-        });
-
-        return cast(parseDateFields(pluckData(response.data)));
-    }
-
     protected async _batchAddRequestsWithRetries(
         requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
         options: RequestQueueClientBatchAddRequestOptions = {},
@@ -148,7 +123,7 @@ export class RequestQueueClient extends ResourceClient {
         // The requests we have not been able to process in the last call
         // ie. those we have not been able to process at all
         let unprocessedRequests: Array<UnprocessedRequest> = [];
-        for (let i = 0; i < maxRetries; i++) {
+        for (let i = 0; i < 1 + maxRetries; i++) {
             const response = await this.httpClient.call({
                 url: this._url('requests/batch'),
                 method: 'POST',
@@ -183,7 +158,7 @@ export class RequestQueueClient extends ResourceClient {
         return cast(parseDateFields(result));
     }
 
-    async batchAddRequestsConcurrently(
+    async batchAddRequests(
         requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
         options: RequestQueueClientBatchAddRequestOptions = {},
     ): Promise<RequestQueueClientBatchAddRequestsResult> {
@@ -195,7 +170,7 @@ export class RequestQueueClient extends ResourceClient {
             forefront: ow.optional.boolean,
             maxRetries: ow.optional.number.greaterThan(0),
             delayBetweenRetriesMillis: ow.optional.number.greaterThan(0),
-            concurrency: ow.optional.number.greaterThan(0),
+            concurrency: ow.optional.number.greaterThan(0).lessThanOrEqual(50),
         }));
 
         const {
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -1,5 +1,5 @@
 import ow from 'ow';
-import type { JsonValue } from 'type-fest';
+import type { JsonObject } from 'type-fest';
 import { ApifyApiError } from '../apify_api_error';
 import { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceClient } from '../base/resource_client';
@@ -145,8 +145,7 @@ export class RequestQueueClient extends ResourceClient {
             await new Promise((resolve) => setTimeout(resolve, this.httpClient.minDelayBetweenRetriesMillis));
         }
 
-        // TODO: Not sure how to do this properly in ts-friendly way
-        const result = <unknown> { processedRequests, unprocessedRequests } as JsonValue;
+        const result = { processedRequests, unprocessedRequests } as unknown as JsonObject;
 
         return cast(parseDateFields(result));
     }
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -105,7 +105,7 @@ export class RequestQueueClient extends ResourceClient {
         // Keep track of the requests that remain to be processed (in parameter format)
         let remainingRequests = requests;
         // Keep track of the requests that have been processed (in api format)
-        let processedRequests: ProcessedRequest[] = [];
+        const processedRequests: ProcessedRequest[] = [];
         // The requests we have not been able to process in the last call
         // ie. those we have not been able to process at all
         let unprocessedRequests: UnprocessedRequest[] = [];
@@ -121,7 +121,7 @@ export class RequestQueueClient extends ResourceClient {
                     }),
                 });
                 const data = response.data.data as RequestQueueClientBatchAddRequestsResult;
-                processedRequests = [...processedRequests, ...data.processedRequests];
+                processedRequests.push(...data.processedRequests);
                 unprocessedRequests = data.unprocessedRequests;
 
                 // Get unique keys of all requests processed so far
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -107,14 +107,7 @@ export class RequestQueueClient extends ResourceClient {
 
         ow(options, ow.object.partialShape({
             forefront: ow.optional.boolean,
-            maxRetries: ow.optional.number.greaterThan(0),
-            delayBetweenRetriesMillis: ow.optional.number.greaterThan(0),
         }));
-
-        const {
-            maxRetries = 9,
-            delayBetweenRetriesMillis = 500,
-        } = options;
 
         // Keep track of the requests that remain to be processed (in parameter format)
         let remainingRequests = requests;
@@ -123,7 +116,7 @@ export class RequestQueueClient extends ResourceClient {
         // The requests we have not been able to process in the last call
         // ie. those we have not been able to process at all
         let unprocessedRequests: Array<UnprocessedRequest> = [];
-        for (let i = 0; i < 1 + maxRetries; i++) {
+        for (let i = 0; i < 1 + this.httpClient.maxRetries; i++) {
             const response = await this.httpClient.call({
                 url: this._url('requests/batch'),
                 method: 'POST',
@@ -149,7 +142,7 @@ export class RequestQueueClient extends ResourceClient {
             }
 
             // Sleep for some time before trying again
-            await new Promise((resolve) => setTimeout(resolve, delayBetweenRetriesMillis));
+            await new Promise((resolve) => setTimeout(resolve, this.httpClient.minDelayBetweenRetriesMillis));
         }
 
         // TODO: Not sure how to do this properly in ts-friendly way
@@ -168,23 +161,19 @@ export class RequestQueueClient extends ResourceClient {
 
         ow(options, ow.object.exactShape({
             forefront: ow.optional.boolean,
-            maxRetries: ow.optional.number.greaterThan(0),
-            delayBetweenRetriesMillis: ow.optional.number.greaterThan(0),
-            concurrency: ow.optional.number.greaterThan(0).lessThanOrEqual(50),
         }));
 
-        const {
-            concurrency = 5,
-        } = options;
+        // The number of 50 parallel requests seemed optimal, if it was higher it did not seem to bring any extra value.
+        const maxParallelRequests = 50;
 
         const operationsInProgress = [];
         const individualResults = [];
 
-        // Send up to `concurrency` requests at once, wait for all of them to finish and repeat
+        // Send up to `maxParallelRequests` requests at once, wait for all of them to finish and repeat
         for (let i = 0; i < requests.length; i += 25) {
             const requestsInBatch = requests.slice(i, i + 25);
             operationsInProgress.push(this._batchAddRequestsWithRetries(requestsInBatch, options));
-            if (operationsInProgress.length === concurrency) {
+            if (operationsInProgress.length === maxParallelRequests) {
                 individualResults.push(...(await Promise.all(operationsInProgress)));
                 operationsInProgress.splice(0, operationsInProgress.length);
             }
@@ -323,9 +312,6 @@ export interface RequestQueueClientAddRequestOptions {
 
 export interface RequestQueueClientBatchAddRequestOptions {
     forefront?: boolean;
-    maxRetries?: number;
-    delayBetweenRetriesMillis?: number;
-    concurrency?: number;
 }
 
 export interface RequestQueueClientRequestSchema {
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -109,6 +109,7 @@ export class RequestQueueClient extends ResourceClient {
         // The requests we have not been able to process in the last call
         // ie. those we have not been able to process at all
         let unprocessedRequests: UnprocessedRequest[] = [];
+        // TODO: Do not use `maxRetries` from http client, but rather a separate value
         for (let i = 0; i < 1 + this.httpClient.maxRetries; i++) {
             try {
                 const response = await this.httpClient.call({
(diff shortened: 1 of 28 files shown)


src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,138 @@ export class RequestQueueClient extends ResourceClient {
...
+        })).minLength(1).maxLength(25));
+
+        ow(options, ow.object.partialShape({
+            forefront: ow.optional.boolean,
+            maxRetries: ow.optional.number.greaterThan(0),
+            delayBetweenRetriesMillis: ow.optional.number.greaterThan(0),
+        }));
+
+        const {
+            maxRetries = 9,
```
fnesveda, line 140: I think the semantic for `maxRetries` is different than for the `maxRetries` in the main client class, which can be confusing - here it's the max number of attempts, but there it is actually the max number of retries (the main client will always make 1 attempt + up to `maxRetries` retries). Could you unify it so it works the same?

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,138 @@ export class RequestQueueClient extends ResourceClient {
...
+            await new Promise((resolve) => setTimeout(resolve, delayBetweenRetriesMillis));
+        }
+
+        // TODO: Not sure how to do this properly in ts-friendly way
+        const result = <unknown> { processedRequests, unprocessedRequests } as JsonValue;
+
+        return cast(parseDateFields(result));
+    }
+
+    async batchAddRequestsConcurrently(
```
fnesveda, line 186: I think the naming here is really confusing, it's not clear why does `batchAddRequests` not do retries, but `batchAddRequestsConcurrently` does them. I'd say we should keep it simple and just have one function, with some sane default concurrency and max retry count, to not confuse users.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,138 @@ export class RequestQueueClient extends ResourceClient {
...
+    ): Promise<RequestQueueClientBatchAddRequestsResult> {
+        ow(requests, ow.array.ofType(ow.object.partialShape({
+            id: ow.undefined,
+        })).minLength(1));
+
+        ow(options, ow.object.exactShape({
+            forefront: ow.optional.boolean,
+            maxRetries: ow.optional.number.greaterThan(0),
+            delayBetweenRetriesMillis: ow.optional.number.greaterThan(0),
+            concurrency: ow.optional.number.greaterThan(0),
```
fnesveda, line 198: Should we set some upper limit for concurrency here? I think the users could quite easily shoot themselves in the foot, with high concurrency, retries for unprocessed items and retries for failed requests (even with exponential backoff), they could be hitting the API rate limit and even retries with exponential backoff could not be enough to prevent it, and the whole request would error out without the user knowing which requests were added and which were not.

Not sure if it would be a common problem, though, or if it would actually matter, they can just retry adding all the requests on error, and the ones which are already in the queue will just not get added. Just a thought for discussion.

fnesveda (changes requested):

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,138 @@ export class RequestQueueClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    async batchAddRequests(
```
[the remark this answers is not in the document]
valekjo: We do not want to include the methods yet, I think not adding any JSDoc is the only way?

src/resource_clients/request_queue.ts
valekjo: I've set the max concurrency to 50, that seemed to perform well in load testing.

src/resource_clients/request_queue.ts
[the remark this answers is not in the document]
fnesveda: Ah, okay, that makes sense :)

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,113 @@ export class RequestQueueClient extends ResourceClient {
...
+        requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
+        options: RequestQueueClientBatchAddRequestOptions = {},
+    ): Promise<RequestQueueClientBatchAddRequestsResult> {
+        ow(requests, ow.array.ofType(ow.object.partialShape({
+            id: ow.undefined,
+        })).minLength(1).maxLength(25));
+
+        ow(options, ow.object.partialShape({
+            forefront: ow.optional.boolean,
+            maxRetries: ow.optional.number.greaterThan(0),
```
fnesveda, line 110: Now with the updated `maxRetries` logic, we could possibly allow setting `maxRetries` to 0 as well.

fnesveda (changes requested):
Looks good, just one last small comment.

src/resource_clients/request_queue.ts
metalwarrior665: Definitely, the client already does things that make life easier and the advanced function does what anyone would do manually anyway.

mnmkng (commented):
Code wise it looks good, but I don't understand why concurrency, max retries or delay should be configurable? Is there any reason why a user would want to add the requests slower than at max speed? I think those options do not bring any value really. Max retries and delays should use the client level values IMO and concurrency should be managed automatically based on rate limit errors from the API.

valekjo:
> Code wise it looks good, but I don't understand why concurrency, max retries or delay should be configurable? Is there any reason why a user would want to add the requests slower than at max speed? I think those options do not bring any value really. Max retries and delays should use the client level values IMO and concurrency should be managed automatically based on rate limit errors from the API.

I have taken max retries and delays from client as suggested. As for the concurrency, I renamed it to `maxParallelRequests` and hardcoded it to 50, which seemed like the highest reasonable value in load testing. If it would in the end be too high in some cases and rateLimit was reached, httpClient would deal with that with it's own retries.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+        requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
+        options: RequestQueueClientBatchAddRequestOptions = {},
+    ): Promise<RequestQueueClientBatchAddRequestsResult> {
+        ow(requests, ow.array.ofType(ow.object.partialShape({
+            id: ow.undefined,
+        })).minLength(1).maxLength(25));
+
+        ow(options, ow.object.partialShape({
+            forefront: ow.optional.boolean,
+        }));
```
mnmkng, line 110: Nit, private methods don't need argument validation.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+
+        // Keep track of the requests that remain to be processed (in parameter format)
+        let remainingRequests = requests;
+        // Keep track of the requests that have been processed (in api format)
+        let processedRequests: Array<ProcessedRequest> = [];
+        // The requests we have not been able to process in the last call
+        // ie. those we have not been able to process at all
+        let unprocessedRequests: Array<UnprocessedRequest> = [];
+        for (let i = 0; i < 1 + this.httpClient.maxRetries; i++) {
+            const response = await this.httpClient.call({
```
mnmkng, line 120: Those two layers of retries still confuse me. So there are retries with exponential backoff from the client, but also retries if all the requests were not added successfully, without exponential backoff, right? Which in turn can trigger more retries with exponential backoff.

I'm not sure this will perform optimally under high load. It could trigger hundreds of retries when the rate limits are hit.

Why would some requests be unprocessed? Because the API is under load and can't save the requests fast enough?

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+        const operationsInProgress = [];
+        const individualResults = [];
+
+        // Send up to `maxParallelRequests` requests at once, wait for all of them to finish and repeat
+        for (let i = 0; i < requests.length; i += 25) {
+            const requestsInBatch = requests.slice(i, i + 25);
+            operationsInProgress.push(this._batchAddRequestsWithRetries(requestsInBatch, options));
+            if (operationsInProgress.length === maxParallelRequests) {
+                individualResults.push(...(await Promise.all(operationsInProgress)));
+                operationsInProgress.splice(0, operationsInProgress.length);
```
mnmkng, line 178: This seems that if any of the batches would throw, the whole thing would throw. But in background, more batches will still be running with unknown states.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+            }
+        }
+        // Get results from remaining operations
+        individualResults.push(...(await Promise.all(operationsInProgress)));
```
mnmkng, line 182: Same here.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+            if (remainingRequests.length === 0) {
+                break;
+            }
+
+            // Sleep for some time before trying again
+            await new Promise((resolve) => setTimeout(resolve, this.httpClient.minDelayBetweenRetriesMillis));
+        }
+
+        // TODO: Not sure how to do this properly in ts-friendly way
+        const result = <unknown> { processedRequests, unprocessedRequests } as JsonValue;
```
mnmkng, line 149: @vladfrangu @B4nan could you please take a look?

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+            });
+            const data = response.data.data as RequestQueueClientBatchAddRequestsResult;
+            processedRequests = [...processedRequests, ...data.processedRequests];
+            unprocessedRequests = data.unprocessedRequests;
+
+            // Get unique keys of all requests processed so far
+            const processedRequestsUniqueKeys = processedRequests.map(({ uniqueKey }) => uniqueKey);
+            // Requests remaining to be processed are the all that remain
+            // @ts-expect-error Undefined id - it seems to be added by ow validation
+            remainingRequests = requests.filter(({ uniqueKey }) => !processedRequestsUniqueKeys.includes(uniqueKey));
```
mnmkng, line 137: What's the difference between `unprocessedRequests` and `remainingRequests`? I thought the unprocessed are being retried but now I see it's the diff between remaining and processed.

mnmkng (changes requested):

src/resource_clients/request_queue.ts
B4nan: i'd not mix the two syntaxes for type assertions, it looks weird. just do `... as unknown as JsonValue`. sometimes there is no better way than casting to unknown/any first

or looking at how the `JsonValue` is defined, it might as well work to type this to `JsonObject` which is part of the `JsonValue` union

src/resource_clients/request_queue.ts
valekjo: The unprocessed requests originate in dynamodb. It happens rarely, but sometimes dynamo is not able to save all requests in the batch. We do retries to dynamo with exponential backoff, but it still might happen that we won't be able to store all the requests. Those are sent back as unprocessed (that's in the design of request queue v2, the other option would be treat the whole batch as failed).

On client, there are indeed two levels of retries:
 - the rate limit retries of the client - those do not care about processed / unprocessed items, and the client takes care of exponential backoff
 - the retries for just the unprocessed items returned from api. I thought it would not be needed to have exponential backoff there, just a small delay if any items remain unprocessed

src/resource_clients/request_queue.ts
valekjo: The `unprocessedRequests` are the ones returned from API. But they do not contain all the fields of request, just `uniqueKey`, `method` and `url`.
`remainingRequests` are in the end the same requests, but with all the fields. It just felt more natural to me to see them as the difference between all requests (`requests`) all requests processed so far (`processedRequests`) instead of just unprocesed items from a single API call.

src/resource_clients/request_queue.ts
valekjo: Thanks.

src/resource_clients/request_queue.ts
```diff
@@ -234,6 +335,24 @@ export interface RequestQueueClientAddRequestResult {
...
+
+interface UnprocessedRequest {
+    uniqueKey: string;
+    url: string;
+    method?: AllowedHttpMethods;
+}
+
+export interface RequestQueueClientBatchAddRequestsResult {
+    processedRequests: Array<ProcessedRequest>,
+    unprocessedRequests: Array<UnprocessedRequest>,
```
B4nan, line 343: we use `[]` syntax for arrays, will add that to the linter

```suggestion
    processedRequests: ProcessedRequest[],
    unprocessedRequests: UnprocessedRequest[],
```

src/resource_clients/request_queue.ts
valekjo: Did not think of this. Maybe it could be dealt with in `_batchAddRequestsWithRetries`, make it never throwing and just take request processed so far into the account.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +97,102 @@ export class RequestQueueClient extends ResourceClient {
...
+        }
+
+        // TODO: Not sure how to do this properly in ts-friendly way
+        const result = <unknown> { processedRequests, unprocessedRequests } as JsonValue;
+
+        return cast(parseDateFields(result));
+    }
+
+    async batchAddRequests(
+        requests: Array<Omit<RequestQueueClientRequestSchema, 'id'>>,
```
B4nan, line 145:
```suggestion
        requests: Omit<RequestQueueClientRequestSchema, 'id'>[],
```

B4nan (changes requested):

fnesveda:
I'm wondering if we should maybe implement this that the server returns a HTTP 207 Multi-Status response, to have it at least a bit standardized.

We'll have to deal with issues like this quite often in the future as we implement more and more batch endpoints, so it would be good to make a robust and easy to reuse solution now (both on the server and in the client), so that we don't have to deal with it later on.

valekjo:
> I'm wondering if we should maybe implement this that the server returns a HTTP 207 Multi-Status response, to have it at least a bit standardized.
> 
> We'll have to deal with issues like this quite often in the future as we implement more and more batch endpoints, so it would be good to make a robust and easy to reuse solution now (both on the server and in the client), so that we don't have to deal with it later on.

After talking to @mtrunkat we'd like to do this as-is for now and check if the batch insert would work fine. We will probably do few more iterations server side.

Adding 207 response should be ok, it should not influence client behavior, will prepare a PR.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +98,108 @@ export class RequestQueueClient extends ResourceClient {
...
+                    url: this._url('requests/batch'),
+                    method: 'POST',
+                    data: remainingRequests,
+                    params: this._params({
+                        forefront: options.forefront,
+                        clientKey: this.clientKey,
+                    }),
+                });
+                const data = response.data.data as RequestQueueClientBatchAddRequestsResult;
+                processedRequests = [...processedRequests, ...data.processedRequests];
```
B4nan, line 124: is there any reason to create new arrays here instead of just pushing to existing one? destructing may look nice, but it will eat more memory than necessary if we are fine with mutating the array, as you are creating new one and pushing all existing items to it first - things you don't with simple `push`

```suggestion
                processedRequests.push(...data.processedRequests);
```

src/resource_clients/request_queue.ts
```diff
@@ -234,6 +342,24 @@ export interface RequestQueueClientAddRequestResult {
...
+    requestId: string;
+    wasAlreadyPresent: boolean;
+    wasAlreadyHandled: boolean;
+}
+
+interface UnprocessedRequest {
+    uniqueKey: string;
+    url: string;
+    method?: AllowedHttpMethods;
+}
```
B4nan, line 356: it feels a bit weird to have two separate interfaces for this, I also don't like the `UnprocessedRequest` name very much - it is a regular request, right? and the processed one is its extension, or not?

```ts
interface ApifyRequest {
    uniqueKey: string;
    url: string;
    method?: AllowedHttpMethods;
}

interface ProcessedRequest extends ApifyRequest {
    requestId: string;
    wasAlreadyPresent: boolean;
    wasAlreadyHandled: boolean;
}
```

seeing the rest of the interfaces here, many others should extend such base request type. it does not have to happen in this PR, just an idea for future.

mnmkng (approved):
Besides the comments from others and for experimental use, I think it's fine.

B4nan:
What's the plan with this PR? I wanted to ship stable client 2.0 and use that in the next SDK version, but was waiting for this PR to be merged. We can ship it as 2.1 once ready, if it might take too long, but there is not much sense to do that if it would be merged later this week.

valekjo:
> What's the plan with this PR? I wanted to ship stable client 2.0 and use that in the next SDK version, but was waiting for this PR to be merged. We can ship it as 2.1 once ready, if it might take too long, but there is not much sense to do that if it would be merged later this week.

@B4nan We have a meeting about this later today, should be clear after.

src/resource_clients/request_queue.ts
valekjo: @B4nan Not really, they just share `uniqueKey` and not the other fields. But we might return those too from API in the future. Is it ok just to add a comment that we should revisit the shapes before we switch from experimental mode?

B4nan (approved):

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +98,111 @@ export class RequestQueueClient extends ResourceClient {
...
+        options: RequestQueueClientBatchAddRequestOptions = {},
+    ): Promise<RequestQueueClientBatchAddRequestsResult> {
+        // Keep track of the requests that remain to be processed (in parameter format)
+        let remainingRequests = requests;
+        // Keep track of the requests that have been processed (in api format)
+        const processedRequests: ProcessedRequest[] = [];
+        // The requests we have not been able to process in the last call
+        // ie. those we have not been able to process at all
+        let unprocessedRequests: UnprocessedRequest[] = [];
+        for (let i = 0; i < 1 + this.httpClient.maxRetries; i++) {
```
drobnikj, line 112: I would do not mix maxRetries with retries for request queue unprocessed items.
Users can set this opt on the client level and probably do not expect this side-effect.

src/resource_clients/request_queue.ts
```diff
@@ -96,6 +98,111 @@ export class RequestQueueClient extends ResourceClient {
...
+                const processedRequestsUniqueKeys = processedRequests.map(({ uniqueKey }) => uniqueKey);
+                unprocessedRequests = requests
+                    .filter(({ uniqueKey }) => !processedRequestsUniqueKeys.includes(uniqueKey))
+                    .map(({ method, uniqueKey, url }) => ({ method, uniqueKey, url }));
+
+                break;
+            }
+
+            // Sleep for some time before trying again
+            await new Promise((resolve) => setTimeout(resolve, this.httpClient.minDelayBetweenRetriesMillis));
```
drobnikj, line 149: Same note as for maxRetries

drobnikj (approved):
Just nits, let's fix them in the next iteration.

src/resource_clients/request_queue.ts
valekjo: Ok, added both as todos.

PR #212, opened 2021-09-28
feat: Adding x-apify-workflow-key header

mtrunkat:
We need to pass this header from the worker env variable in order to be able to associate runs spun as a part of one flow.


B4nan (approved):

valekjo (approved):

B4nan:
Planning to ship stable 2.0 of client today and use that in the SDK, I guess we can merge this, right @mtrunkat?

PR #214, opened 2021-11-17
chore: make Datasets support generics

vladfrangu:
For https://github.com/apify/apify-ts/pull/4

### DIFF
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -136,7 +136,9 @@ export class ApifyClient {
     /**
      * https://docs.apify.com/api/v2#/reference/datasets/dataset
      */
-    dataset(id: string): DatasetClient {
+    dataset<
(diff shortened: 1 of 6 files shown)


vladfrangu:
CI is running on nodejs 14 for some reason... shouldn't that be removed?

mnmkng:
> CI is running on nodejs 14 for some reason... shouldn't that be removed?

The SDK supports only Node 15+, but there's no reason for the client to only support those versions. People might want to use it in various applications.

vladfrangu:
> > CI is running on nodejs 14 for some reason... shouldn't that be removed?
> 
> The SDK supports only Node 15+, but there's no reason for the client to only support those versions. People might want to use it in various applications.

👍
Found and fixed the issue with CI failing too

B4nan (approved):

PR #218, opened 2021-12-09
chore: fix some annoying types

vladfrangu:
While working on https://github.com/apify/apify-ts/pull/11, I noticed a lot of duplicated types that can be united and re-used in the SDK
This PR corrects some of them. I will submit more PRs as I find them necessary


B4nan (approved):
Perfect, indeed there will need to be adjustments in other repositories, not just client (I see you browser pool).

PR #219, opened 2021-12-14
fix: unnamed storages

mnmkng:
Thanks to a change in API, we could not create unnamed storages with the Client anymore. This fixes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+2.0.4/ 2021/12/XX
+===================
+- Fix: Unnamed storages can now be created again.
+
 2.0.3 / 2021/11/18
 ===================
 - For TypeScript users, DatasetClients can now take in a generic type parameter that defines the data present in a dataset.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "apify-client",
-	"version": "2.0.4",
+	"version": "2.0.5",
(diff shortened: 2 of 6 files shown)


fnesveda (approved):
The PR looks good, but I'm curious - do you know when did it break? I checked the API code and I don't think we made any changes to the handling of the `name` parameter for at least a year.

B4nan:
My guess is that this sneaked in during the TS rewrite, so with v2 of apify-client.

B4nan:
mmm not really, v1.4 also uses the default parameter

https://github.com/apify/apify-client-js/blob/v1.4.2/src/base/resource_collection_client.js#L47

mnmkng:
> The PR looks good, but I'm curious - do you know when did it break? I checked the API code and I don't think we made any changes to the handling of the `name` parameter for at least a year.

I don't know when it happened, but we even talked about it with @mtrunkat. Basically, I found that you can create unnamed storages when you pass `name=` as a param and we were able to simulate this in the client using `''`. With @mtrunkat we agreed that it makes sense to allow users to create unnamed storages that will be automatically deleted after data retention period. But the API was not ideal, so we agreed that it would make better sense to just create it when you don't pass any name.

Apparently it already happened. So `name=` does not work anymore and throws an error that `name must be a string`. Whereas completely removing the name correctly creates an unnamed dataset now.

PR #221, opened 2022-01-03
fix: correct type for webhook event type

vladfrangu:


B4nan (approved):

B4nan:
we will need a version bump + a note in changelog

PR #222, opened 2022-01-03
fix: correct validation for Task call & start, and loosen up KeyValueStore getRecord options

vladfrangu:
I'm not entirely sure if this is a breaking change or not, so I'm awaiting your thoughts on it. This PR aligns types closer to what the API wants as well as relaxes the types for KeyValueStore#getRecord.

I have the changelog prepared, just have to change the version number once I know which is the correct one to change it to

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -5,6 +5,7 @@ import { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceClient } from '../base/resource_client';
 import {
     cast,
+    Dictionary,
     parseDateFields,
     pluckData,
     stringifyWebhooksToBase64,
@@ -56,8 +57,8 @@ export class ActorClient extends ResourceClient {
      * Starts an actor and immediately returns the Run object.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
-    async start(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
-        // input can be anything, pointless to validate
+    async start(input?: string | Dictionary, options: ActorStartOptions = {}): Promise<ActorRun> {
+        ow(input, ow.optional.any(ow.object, ow.string));
         ow(options, ow.object.exactShape({
             build: ow.optional.string,
             contentType: ow.optional.string,
@@ -104,8 +105,8 @@ export class ActorClient extends ResourceClient {
      * It waits indefinitely, unless the `waitSecs` option is provided.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
-    async call(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
-        // input can be anything, pointless to validate
+    async call(input?: string | Dictionary, options: ActorStartOptions = {}): Promise<ActorRun> {
+        ow(input, ow.optional.any(ow.object, ow.string));
         ow(options, ow.object.exactShape({
             build: ow.optional.string,
             contentType: ow.optional.string,
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -5,7 +5,6 @@ import { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceClient } from '../base/resource_client';
 import {
     cast,
-    Dictionary,
     parseDateFields,
     pluckData,
     stringifyWebhooksToBase64,
@@ -57,8 +56,7 @@ export class ActorClient extends ResourceClient {
      * Starts an actor and immediately returns the Run object.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
-    async start(input?: string | Dictionary, options: ActorStartOptions = {}): Promise<ActorRun> {
-        ow(input, ow.optional.any(ow.object, ow.string));
+    async start(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
         ow(options, ow.object.exactShape({
             build: ow.optional.string,
             contentType: ow.optional.string,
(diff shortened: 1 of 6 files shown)


src/resource_clients/key_value_store.ts
```diff
@@ -78,9 +78,15 @@ export class KeyValueStoreClient extends ResourceClient {
      */
     async getRecord(key: string): Promise<KeyValueStoreRecord<JsonValue> | undefined>;
 
-    async getRecord(key: string, options: { buffer: true }): Promise<KeyValueStoreRecord<Buffer> | undefined>;
-
-    async getRecord(key: string, options: { stream: true }): Promise<KeyValueStoreRecord<ReadableStream> | undefined>;
+    async getRecord<Options extends KeyValueClientGetRecordOptions>(
+        key: string,
+        options: Options
+    ): Promise<KeyValueStoreRecord<Options['stream'] extends true ? ReadableStream : JsonValue> | undefined>;
```
B4nan, line 84: let's extract the conditional mapped type to a type alias

B4nan (changes requested):
code wise looking good, for me this is a bug fix, so should be a patch bump 

if we don't have such tests already, we should add some that are using the string parameter (i guess we don't as it was disallowed via `ow` checks on some places)

vladfrangu:
> if we don't have such tests already, we should add some that are using the string parameter
We already have tests for those, we just didn't have validation for the input

B4nan:
this method had the validation and did not allow strings

https://github.com/apify/apify-client-js/pull/222/files#diff-fa18a287becfdb48f13b30ce26c2c506878641aaca2c7ab9b1b5fd2f7a75470dL59

src/resource_clients/actor.ts
```diff
@@ -56,8 +57,8 @@ export class ActorClient extends ResourceClient {
      * Starts an actor and immediately returns the Run object.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
-    async start(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
-        // input can be anything, pointless to validate
+    async start(input?: string | Dictionary, options: ActorStartOptions = {}): Promise<ActorRun> {
+        ow(input, ow.optional.any(ow.object, ow.string));
```
mnmkng, line 61: Actor input can be anything. This is a breaking change imo.

src/resource_clients/actor.ts
```diff
@@ -104,8 +105,8 @@ export class ActorClient extends ResourceClient {
      * It waits indefinitely, unless the `waitSecs` option is provided.
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
-    async call(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
-        // input can be anything, pointless to validate
+    async call(input?: string | Dictionary, options: ActorStartOptions = {}): Promise<ActorRun> {
+        ow(input, ow.optional.any(ow.object, ow.string));
```
mnmkng, line 109: ditto

src/resource_clients/task.ts
```diff
@@ -55,8 +55,8 @@ export class TaskClient extends ResourceClient {
      * Starts a task and immediately returns the Run object.
      * https://docs.apify.com/api/v2#/reference/actor-tasks/run-collection/run-task
      */
-    async start(input: Dictionary | Dictionary[], options: TaskStartOptions = {}): Promise<ActorRun> {
-        ow(input, ow.optional.object);
+    async start(input: string | Dictionary, options: TaskStartOptions = {}): Promise<ActorRun> {
+        ow(input, ow.optional.any(ow.object, ow.string));
```
mnmkng, line 59: But it needs to be a JSON string. That's why we kept object, to make sure that it's JSON serializable before it hits API.

src/resource_clients/actor.ts
vladfrangu: The API begs to differ, and rejects anything that isn't an object, so shouldn't we also validate for that?

vladfrangu: > btw the type accepts also `undefined`, which would fail this runtime validation

It wouldn't fail the runtime validation, since the entire predicate is optional (tested this too)

mnmkng: The API will accept whatever you set the `content-type` to. E.g. if you set it to `application/pdf` then it will process it as a buffer.

This does not apply to tasks. Tasks accept only JSON input.

vladfrangu: Ah, its a case of Actor runs vs Tasks.. Will revert tomorrow 👍

src/resource_clients/task.ts
vladfrangu: Could validate that the string is a valid JSON string? Would that be better, or should we just stick to allowing just objects?

src/resource_clients/actor.ts
```diff
@@ -57,7 +57,6 @@ export class ActorClient extends ResourceClient {
      * https://docs.apify.com/api/v2#/reference/actors/run-collection/run-actor
      */
     async start(input?: unknown, options: ActorStartOptions = {}): Promise<ActorRun> {
-        // input can be anything, pointless to validate
```
B4nan, line 60: > The API will accept whatever you set the content-type to. E.g. if you set it to application/pdf then it will process it as a buffer.

would be good to note this in the comment instead of removing it, same for line 108

src/resource_clients/task.ts
mnmkng: Yeah, could be. I guess I was just lazy when I was making those validations. Thinking that the number of users who would want to use JSON string as an input instead of an object would be really low.

vladfrangu: What do you suggest I do? make a validator that makes sure the string is a valid json object, or leave it as is?

B4nan: How does it behave now? I guess in case of invalid JSON there will be some error when we call `JSON.parse` anyway? Which seems enough to me, so I would keep it as is in such case.

If we would validate this, we should not parse the JSON string twice (once for validation and once for actual payload) to mitigate possible performance issues.

vladfrangu: Where do we call `JSON.parse` exactly? It seems we only do that for the response, not the request...unless axios does it by default?

mnmkng: I would just keep it as object. We can allow string if some users request, but it's super easy for them to parse their own strings into objects, whereas for us, it's a bit of a pain to do the checks. And also I'm pretty sure that if we allow type string, people will start putting random strings there.

B4nan (approved):

PR #230, opened 2022-02-09
fix: fields ending in `At` are Date objects

### LINKED ISSUE #228: Actor run date fields of type string
CleanShot 2022-02-04 at 17 23 15@2x

vladfrangu:
Closes #228


B4nan (approved):
looking good, just add a note to the changelog

vladfrangu:
Does this count as a patch release?

B4nan:
doesnt matter (but yes, it would be), we alread have a feat commit and 2.2 is not yet release, so it will be part of it

Issue #229, opened 2022-02-08
Typings for 2.1.0. are off


mhamas:
When trying 
```
await this.apifyClient!
    .tasks()
    .create({
        actorId,
        name: taskName,
        options: { build: buildNumber, timeout: TASK_TIMEOUT_SECONDS },
    }); 
```
I run into the following issues
- `actorId` is not present in typings and it seems that the client actually expects `actId`
- `timeout` is present in typings but the runtime complains. it seems the legacy `timeoutSecs` works

vladfrangu:
Looking at the code, seems like the only properties that cannot be changed are `userId`, `stats` and the `*At` fields. If that's correct I'll update the types accordingly.

For the `timeout` vs `timeoutSecs`, from what I see TaskOptions has only `timeoutSecs` defined. 👀 Is timeoutSecs deprecated? If so, I can update it 👍

mhamas:
Hm...looking at the API probably not and it should e still `timeoutSecs`. But I believe the type hints in IDE were showing me `timeout` instead.

vladfrangu:
> But I believe the type hints in IDE were showing me `timeout` instead

What version of the client are you on? It looks fine on 2.1.0 too 👀 
![image](https://user-images.githubusercontent.com/17960496/153392887-4dd1ef43-664e-4beb-9251-9e6b81cef49d.png)

mhamas:
Ok, this is probably my bad. Checking it now and it does show `timeoutSecs`.

vladfrangu:
I'm gonna close this for cleanup sake, please re-open or followup if this is still an issue 🙏

PR #231, opened 2022-02-09
fix(typings): correct types and validation for `tasks().create`

### LINKED ISSUE #227: Task update data miss actorId type
image

vladfrangu:
Closes #227, and resolves one of the issues mentioned in #229 

Awaiting response in Slack to see if the actId should also be allowed in TaskUpdateData or not

### DIFF
--- a/src/resource_clients/task_collection.ts
+++ b/src/resource_clients/task_collection.ts
@@ -36,8 +36,8 @@ export class TaskCollectionClient extends ResourceCollectionClient {
     /**
      * https://docs.apify.com/api/v2#/reference/actor-tasks/task-collection/create-task
      */
-    async create(task?: TaskUpdateData): Promise<Task> {
-        ow(task, ow.optional.object);
+    async create(task: TaskCreateData): Promise<Task> {
+        ow(task, ow.object);
 
         return this._create(task);
     }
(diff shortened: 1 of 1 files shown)


mhamas (approved):
Looks good to me, but I'm not honestly sure about the exact shape of the data that should be present.

vladfrangu:
> Looks good to me, but I'm not honestly sure about the exact shape of the data that should be present.

Seems like the API requires only the actId to create a task, and the other properties are optional. I should probably make validation stricter for that, but I'll wait for @B4nan's thoughts

B4nan:
You mean runtime validation via `ow`? Yeah let's align that with how TS checks it.

mnmkng:
> You mean runtime validation via ow? Yeah let's align that with how TS checks it.

I think originally, we left out all runtime validation of API properties, because we did not want to keep those in sync. The users will get a reasonable error from the API if they pass wrong options. We understood that we could save ourselves the request, but figured that was not a sufficient benefit to outweigh the need to keep the API and runtime validations in sync.

But now we have TypeScript so we have to keep it in sync anyway. If you decide to add the runtime validation, please add it everywhere. Would be weird to validate it for update task and not elsewhere.

B4nan:
What about this, I guess we can merge, right?

vladfrangu:
We can merge this as is, and I can do a PR next week to add validation in line with the TS interfaces across the repo

PR #232, opened 2022-02-10
feat: add option to set timeout for request queue client

AndreyBykov:
The default timeout for API client (360 secs) is too high for some endpoints. Add an option to set timeout to request queue separately from the main API client timeout.

### DIFF
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -192,12 +192,11 @@ export class ApifyClient {
         ow(id, ow.string.nonEmpty);
         ow(options, ow.object.exactShape({
             clientKey: ow.optional.string.nonEmpty,
-            timeout: ow.optional.number,
+            timeoutSecs: ow.optional.number,
         }));
 
         const apiClientOptions = {
             id,
-            timeout: options.timeout,
             ...this._options(),
         };
         return new RequestQueueClient(apiClientOptions, options);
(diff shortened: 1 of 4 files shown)


src/apify_client.ts
```diff
@@ -192,10 +192,12 @@ export class ApifyClient {
         ow(id, ow.string.nonEmpty);
         ow(options, ow.object.exactShape({
             clientKey: ow.optional.string.nonEmpty,
+            timeout: ow.optional.number,
         }));
 
         const apiClientOptions = {
             id,
+            timeout: options.timeout,
```
B4nan, line 200: this seems wrong, it should be part of the second parameter only, that is how you use it in `RequestQueueClient`

B4nan (changes requested):
please add tests

mnmkng (commented):
I think the correct place to add this would be either everywhere or to dataset and key-value store clients. Because that is where we need to override the low timeout because of long uploads/downloads. Everywhere else the timeout can be short. I understand that adding the timeout to the queue is the quickest fix, but conceptually it does not make sense.

B4nan:
> I think the correct place to add this would be either everywhere or to dataset and key-value store clients. Because that is where we need to override the low timeout because of long uploads/downloads. Everywhere else the timeout can be short. I understand that adding the timeout to the queue is the quickest fix, but conceptually it does not make sense.

This is supposed to be a hotfix, allowing us to change the long default timeout to 30s in the SDK. I want it this way so we can release it today, without arguing where we need long timeout and what is the right default value for it (and without possibly breaking use cases that need the long timeout).

I believe you did not like the idea of having separate defaults for this, so your suggestion to change deafult to 30s seems quite breaking.

B4nan (approved):
add a note to changelog, otherwise lgtm

(and let's improve on this next week I guess)

B4nan (approved):

PR #233, opened 2022-02-14
feat: Add schema parameter to getOrCreate for kvs, dataset

jirimoravcik:


mtrunkat (approved):

B4nan (approved):

jirimoravcik:
@B4nan Should I bump the package version or something?

B4nan:
right, 2.2.0 is out, so we need a bump 

technically minor one, personally i'd just do a patch as it seems like a small change, will leave it up to you :]

jirimoravcik:
> right, 2.2.0 is out, so we need a bump
> 
> technically minor one, personally i'd just do a patch as it seems like a small change, will leave it up to you :]

Went for 2.2.1, added an entry to the changelog, is it ok like this?

B4nan:
sure, feel free to merge

PR #239, opened 2022-04-07
feat: simplify batch add requests

drobnikj:


mnmkng:
I think we moved past the "thin wrapper" idea when we released version 1. The client should simplify integration of Apify. To me, the function without the retries is a bit useless, because you have to understand API internals very well to be able to implement the retries. What's the use-case for having the non-retrying function?

B4nan:
Agreed with Ondra, from the SDK standpoint, we expect the client to handle retries, we even use it to get the rate limit stats for autoscaling. All the other methods are handling this (and there is no other `*WithRetries` method), why should we have a single outlier that don't. I'd keep just one public method that handles all, number of retries should be configurable from outside - if you pass 0 retries, you have the simplified version without them.

drobnikj:
> What's the use-case for having the non-retrying function?

I have in mind mainly SDK, that it will make sense that SDK will handle autoscaling based on unprocessRequests.

But I got your points that you can handle autoscaling based on a number of retries and if it will be not enough we can introduce a new stat for request queue which SDK can use.

mnmkng:
> SDK will handle autoscaling based on unprocessRequests.

SDK autoscales based on 429 errors from API. If API does not return 429s, SDK will not downscale. I would not introduce additional complexity to this on the SDK level. Either the batch function needs to autoscale on its own or return 429s.

mnmkng:
How do we handle the problem where a user starts pushing a lot of batches to the API and only learns that the API is overloaded 2 minutes later when the exponential backoff on the backend runs out of retries?

I still think that the optimal throughput needs to be handled in the client, based on some immediate information from the API. But even if I ignore that, to be able to properly autoscale on the SDK level, we need to know that the API is overloaded early. After the the two minutes of silence, where the SDK will be scaling up (unless CPU caps it) it could easily send tens of thousands of requests to the API. Then it would get most of them back as unprocessed and the cycle would start over.

Has there been any development in this?

mnmkng:
How do we handle the problem where a user starts pushing a lot of batches to the API and only learns that the API is overloaded 2 minutes later when [the exponential backoff on the backend](https://www.notion.so/apify/Request-queue-v2-cdf4c66e05984214a251fca464695f42#8facb5e13cff45cf9a6bf701f635e112) runs out of retries?

I still think that the optimal throughput needs to be handled in the client, based on some immediate information from the API. But even if I ignore that, to be able to properly autoscale on the SDK level, we need to know that the API is overloaded early. After the the two minutes of silence, where the SDK will be scaling up (unless CPU caps it) it could easily send tens of thousands of requests to the API. Then it would get most of them back as unprocessed which would save to stats as a lot of rate limit errors and after that the process would start over.

fnesveda:
@mnmkng We are passing the message to the SDK early - there are no retries on the server, and the client [increases the `stats.rateLimitErrors`](https://github.com/apify/apify-client-js/pull/239/files#diff-0bb99f12bfadd00c1dd0148f8776d6e2cfa20dabed6e9b52f1207dde1b7303dbR168) after each HTTP call if there are still some unprocessed requests. So the autoscaled pool in SDK will know about the rate limiting immediately when it reads those client stats, which it does once per second.

Conceptually, I agree that it would be nice if the client handled this autoscaling itself, so even other tools than the SDK can benefit from it, but since we have the scaling logic in the SDK, let's use that for now, since the client is used through the SDK the most, by far.

mnmkng:
@fnesveda so just to double check, [this](https://www.notion.so/apify/Request-queue-v2-cdf4c66e05984214a251fca464695f42#8facb5e13cff45cf9a6bf701f635e112) is not true:

> DynamoDB may return a list of unprocessed requests (we need to measure how often) and the server will be automatically retrying those with exponential backoff for up to 2 minutes. After this period if there is anything unprocessed it will be returned to the client to retry it client-side.

fnesveda:
@mnmkng Yup, not anymore, we decided to remove the server-side retries, so that the client gets the rate limit errors as soon as possible. We forgot to remove the mention in Notion, sorry.

src/resource_clients/request_queue.ts
```diff
@@ -103,33 +108,65 @@ export class RequestQueueClient extends ResourceClient {
...
+                const response = await this._batchAddRequests(remainingRequests, {
+                    forefront,
                 });
-                const data = response.data.data as RequestQueueClientBatchAddRequestsResult;
-                processedRequests.push(...data.processedRequests);
-                unprocessedRequests = data.unprocessedRequests;
+                processedRequests.push(...response.processedRequests);
+                unprocessedRequests = response.unprocessedRequests;
+
+                // Consider request with unprocessed requests as rate limited.
```
mtrunkat, line 166: Just add note that this is important to let AutoscalingPool in SDK know that it should potentially downscale.

src/resource_clients/request_queue.ts
```diff
@@ -103,33 +108,65 @@ export class RequestQueueClient extends ResourceClient {
...
         // Keep track of the requests that remain to be processed (in parameter format)
         let remainingRequests = requests;
         // Keep track of the requests that have been processed (in api format)
         const processedRequests: ProcessedRequest[] = [];
         // The requests we have not been able to process in the last call
         // ie. those we have not been able to process at all
         let unprocessedRequests: UnprocessedRequest[] = [];
-        // TODO: Do not use `maxRetries` from http client, but rather a separate value
-        for (let i = 0; i < 1 + this.httpClient.maxRetries; i++) {
+        for (let attempt = 0; attempt < 1 + maxUnprocessedRequestsRetries; attempt++) {
```
mtrunkat, line 158: This is something we should discuss perhaps this value should be much larger not just 8. As in the majority of cases, you want it to keep retrying as long as it's needed although it may take 20mins.

Or we could override this when used from SDK.

CC @B4nan @mnmkng

mtrunkat (changes requested):

src/resource_clients/request_queue.ts
mnmkng: I think in terms of making updates to the queue, we prefer fast info. Some of the most annoying bugs in the SDK lately were caused by request queue getting stuck for minutes. It's one of the causes of the 0 concurrency bug. Anything more than 30 secs is a bad idea imo.

drobnikj: > Anything more than 30 secs is a bad idea imo.

Timeout probably makes sense but 30 secs is too optimistic, as this method should be used to pass thousands of requests into the queue. From tests of batch insert, it looks like we can insert 1k requests per sec(of course in depends on how heavily the request queues are used at the moment.).

mnmkng: Hmm, I'm not entirely sure about it, but I guess I wouldn't want to push one million requests with a single function call that can potentially last 1000 seconds. Maybe we should limit the batch size to 5000 or something? We can always increase the limit if we find ways to make this faster.

mnmkng: cc @B4nan ad above

B4nan: We do override the timeout for request queue client to 30s in SDK (and retry 3 times ourselves), we actually believe that waiting for longer was one of the things behind the 0 concurrency bug. Agreed it makes sense for us to do things in smaller batches like 1000 requests and not wait for the whole import (cc @vladfrangu we should revisit this). But that can be already done, its up to us how many requests we pass to the method, right? I don't mind if the client itself allows unlimited amount, as I am in the control when I use it. We can await the first batch of 1000 and schedule the rest to run in background (but I guess serially so we don't put too much pressure on the API).

drobnikj: @B4nan Yep, you can pass how many requests you want to and maxParallel option is set up to optimal values which come from testing, but you can change it if you see some issues.

PR #243, opened 2022-05-05
fix: Add exponential backoff to batchAddRequests, untie batch and client settings

jirimoravcik:


mtrunkat (approved):

src/resource_clients/request_queue.ts
```diff
@@ -226,9 +231,9 @@ export class RequestQueueClient extends ResourceClient {
         const operationsInProgress = [];
         const individualResults = [];
 
-        // Send up to `maxParallelRequests` requests at once, wait for all of them to finish and repeat
-        for (let i = 0; i < requests.length; i += 25) {
-            const requestsInBatch = requests.slice(i, i + 25);
+        // Send up to `maxParallel` requests at once, wait for all of them to finish and repeat
+        for (let i = 0; i < requests.length; i += MAX_REQUESTS_PER_BATCH_OPERATION) {
+            const requestsInBatch = requests.slice(i, i + MAX_REQUESTS_PER_BATCH_OPERATION);
```
fnesveda, line 236: We could simplify this using `async.mapLimit` and `_.chunk`, the code would be a lot cleaner, and the operations would run with the specified concurrency all the time, instead of just having batches that are processed in parallel.

src/resource_clients/request_queue.ts
```diff
@@ -190,9 +192,12 @@ export class RequestQueueClient extends ResourceClient {
...
             }
 
-            // Sleep for some time before trying again
-            // TODO: Do not use delay from client, but rather a separate value
-            await new Promise((resolve) => setTimeout(resolve, this.httpClient.minDelayBetweenRetriesMillis));
+            // Exponential backoff
+            const delayMillis = Math.max(
+                Math.floor(Math.random() * (2 ** i)) * minDelayBetweenUnprocessedRequestsRetriesMillis,
+                minDelayBetweenUnprocessedRequestsRetriesMillis,
+            );
```
fnesveda, line 199: I think if you have the random factor between 0 and 1, you'll often run into situations where the delay is shorter than it should. By having the random factor between 1 and 2, you can simplify the code a bit, and have more consistently increasing delays:
```suggestion
            const delayMillis = Math.floor((1 + Math.random()) * (2 ** i) * minDelayBetweenUnprocessedRequestsRetriesMillis);
```

fnesveda (changes requested):
Looks good, I just had a few improvement ideas.

src/resource_clients/request_queue.ts
jirimoravcik: I copied it from apify-core batch rq, so I guess it's also wrong there

src/resource_clients/request_queue.ts
jirimoravcik: We would need to add a dependency on `async`, right? I don't really want to do that.

Yeah as said in slack, I'm working on the concurrency, I will add it to this pr

valekjo (approved):
👍

src/resource_clients/request_queue.ts
mnmkng: We're already using https://github.com/vercel/async-retry for the regular exponential backoff so you can use this if it helps.

src/resource_clients/request_queue.ts
mnmkng: I guess it's relevant even for this https://github.com/apify/apify-client-js/pull/243#discussion_r866949635

PR #249, opened 2022-05-20
feat: add request queue requests list method

drobnikj:
* New method to list requests
* New method to easily iterate all requests in queue
Example usage:
```
for await (const { items } of rqClient.paginateRequests()) {
  const rqsToDelete = items.filter(someFilter);
  await rqClient.batchDeleteRequests(rqsToDelete);
}
```

### DIFF
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -464,7 +464,9 @@ export class RequestQueueClient extends ResourceClient {
      * @private
      * @experimental
      */
-    paginateRequests(options: RequestQueueClientPaginateRequestsOptions = {}): RequestQueueRequestsAsyncIterable<RequestQueueClientListRequestsResult> {
+    paginateRequests(
+        options: RequestQueueClientPaginateRequestsOptions = {},
+    ): RequestQueueRequestsAsyncIterable<RequestQueueClientListRequestsResult> {
         ow(options, ow.object.exactShape({
             limit: ow.optional.number,
             maxPageLimit: ow.optional.number,
@@ -530,7 +532,7 @@ export interface RequestQueueClientListRequestsOptions {
     exclusiveStartId?: string;
 }
 
-export interface RequestQueueClientRequestsOptions {
+export interface RequestQueueClientPaginateRequestsOptions {
     limit?: number;
     maxPageLimit?: number;
     exclusiveStartId?: string;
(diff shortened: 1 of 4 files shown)


mtrunkat (approved):

src/resource_clients/request_queue.ts
```diff
@@ -425,6 +427,57 @@ export class RequestQueueClient extends ResourceClient {
...
+     * THIS METHOD IS EXPERIMENTAL AND NOT INTENDED FOR PRODUCTION USE.
+     *
+     * Usage:
+     * for await (const { items } of client.paginateRequests({ limit: 10 })) {
+     *   items.forEach((request) => console.log(request));
+     * }
+     * @private
+     * @experimental
+     */
+    paginateRequests(options: RequestQueueClientRequestsOptions = {}): RequestQueueRequestsAsyncIterable<RequestQueueClientListRequestsResult> {
```
fnesveda, line 467: I think it should be like this:
```suggestion
    paginateRequests(options: RequestQueueClientPaginateRequestsOptions = {}): RequestQueueRequestsAsyncIterable<RequestQueueClientListRequestsResult> {
```

fnesveda (changes requested):
Nice! I love the async iterator over the items, that's something I'd like to have for all the `listSomething` methods in the client, like [OctoKit has it](https://github.com/octokit/plugin-paginate-rest.js/).

Just one idea - maybe it would be better to `yield` the requests from the `AsyncIterator` one by one, instead of as whole pages? That way you'd hide the whole pagination logic from the users completely, and they could just do:
```js
for await (const request of client.requestQueue(queueId).paginateRequests()) {
    // do something with `request`
}
```

I think it wouldn't be a big change, you could just do `yield* page.items` instead of `yield page`.

drobnikj:
> `yield` the requests from the `AsyncIterator` one by one

I had it like that originally, but I found out that with one by one you cannot easily call batch operations like batchDeleteRequests.

mnmkng:
Maybe it could be an option? `yieldRequests: true` or smth like that. I think both use-cases are valid.

drobnikj:
> Maybe it could be an option?

I would not overcomplicate now. I would say the main customer for this is SDK, so we can adjust if we found we need request by request during the implementation of rq2 into SDK.

B4nan (approved):
> so we can adjust if we found we need request by request during the implementation of rq2 into SDK.

I am aligned with this - let's improve when we see its needed.

PR #251, opened 2022-05-25
refactor(docs): use typedoc

### LINKED ISSUE #236: Generating README not working
`npm run build_docs` script that generates the markdown from jsdocs is not working.

There seems to be an issue with `jsdocs-markdown` and `typescript` in general as running the script throws `UnhandledPromiseRejectionWarning: JSDOC_ERROR: There are no input files to process.`. Also since we switched to typescript we also stopped using jsdocs comments.

Right now we have to update the README manually which is not ideal as it can produce issues like -> https://apifier.slack.com/archives/C0L33UM7Z/p1646217864275879

We could possibly use [typedoc package](https://typedoc.org/) along with its [plugin](https://www.npmjs.com/package/typedoc-plugin-markdown) for turning the default html output to markdown

m0uka:
Uses [typedoc](https://typedoc.org/) instead of jsdoc for documentation.
`typedoc-plugin-markdown` is used to generate markdown instead of HTML.

The docs are no longer in README, but in a `docs` folder, and are separated, the README would be too large otherwise (10k+ lines), however a script can be written or [concat-md](https://github.com/ozum/concat-md) can be used to concatenate them.

You can view the docs [here](https://github.com/m0uka/apify-client-js/blob/master/docs/modules.md).

Some files were also edited to conform with typedoc (`@hideconstructor` -> `@ignore`).
HTML-based docs could be also used later - see [example](https://typedoc.org/example/)

Closes #236, tagging @mtrunkat


mnmkng:
Just my 2 cents. The generated docs look OK, but throwing the user into the[ table of contents](https://github.com/apify/apify-client-js/blob/ee65e8e0353d60a071d602f7107d5085cfdc33d9/docs/modules.md) full of classes and interfaces is not great imo. It might be useful for TS users who are already working with the project and need to quickly find docs for something specific, but I assume they would use their IDE functions to do that.

But for all new users who just arrived, know nothing about the structure and just want to browse a bit, the table of contents could be confusing, because it has no notion of what's important, where they should start and what are the classes and functions they will spend 90% of time working with.

The original docs, for example, list all the functions of the ApifyClient right at the top of the API reference. It also does not list any of the TS interfaces and types that a regular user won't need early in the process.

Can we restructure the ToC to make it more helpful and beginner friendly?

B4nan:
I believe we wanted to have a new monorepo with actor sdk + apify client + apify storage local + some other apify related things, and I was thinking about reusing the docusaurus setup we already have for crawlee, including the API docs. I guess it makes more sense for TS projects, as its valid to have all the information there, but its hard to fit it into a single readme file.

mtrunkat:
@B4nan definitely the monorepo but these docs should be part of docs.apify.com (and later moved to developers.apify.com) so we just need to render markdowns/HTML from TS code and then we embed it there.

B4nan:
Got it. But maybe it would be better to have the rendering logic in docs.apify.com instead, and use the `typedoc.json` directly (or have a separate package for this) than doing it directly in client. I guess the same will be needed for Actor SDK, right?

PR #256, opened 2022-06-27
Support streaming request body

mvolfik:
As discussed in https://apifier.slack.com/archives/CD0SF6KD4/p1655891141193099

I'm still not completely convinced it is ok to switch the default behavior like this - maybe a warning could be issued for a few versions when a stream is passed for the first time during a run

Also any ideas on how to test this are welcome, but what I was doing so far is just run it with a stream of a biig (500 MB+) file and then observe memory usage of the `node` process

### DIFF
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -23,8 +23,6 @@ const { version } = dynamicRequire('../package.json');
 
 const RATE_LIMIT_EXCEEDED_STATUS_CODE = 429;
 
-let warnedAboutStreamNoRetry = false;
-
 export class HttpClient {
     stats: Statistics;
 
@@ -134,12 +132,8 @@ export class HttpClient {
     }
 
     private _informAboutStreamNoRetry() {
-        // using a module-global variable to only warn once per whole process
-        if (!warnedAboutStreamNoRetry) {
-            warnedAboutStreamNoRetry = true;
-            this.logger.warning('Request body was a stream - retrying will not work, as part of it was already consumed.');
-            this.logger.info('If you want the SDK to handle retries, collect the stream into a buffer before sending it.');
-        }
+        this.logger.warning('Request body was a stream - retrying will not work, as part of it was already consumed.');
+        this.logger.info('If you want the SDK to handle retries, collect the stream into a buffer before sending it.');
     }
 
     /**
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -148,6 +148,9 @@ export class HttpClient {
             const requestIsStream = isStream(config.data);
             try {
                 if (requestIsStream) {
+                    // Handling redirects is not possible without buffering - part of the stream has already been sent and can't be recovered
+                    // when server sends the redirect. Therefore we need to override this in Axios config to prevent it from buffering the body.
+                    // see also axios/axios#1045
                     config = { ...config, maxRedirects: 0 };
                 }
                 response = await this.axios.request(config);
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -133,7 +133,7 @@ export class HttpClient {
 
     private _informAboutStreamNoRetry() {
         this.logger.warningOnce('Request body was a stream - retrying will not work, as part of it was already consumed.');
-        this.logger.warningOnce('If you want the SDK to handle retries, collect the stream into a buffer before sending it.');
+        this.logger.warningOnce('If you want Apify client to handle retries for you, collect the stream into a buffer before sending it.');
     }
 
     /**
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -140,8 +140,9 @@ export class HttpClient {
         const makeRequest: RetryFunction<ApifyResponse> = async (stopTrying, attempt) => {
             this.stats.requests++;
             let response: ApifyResponse;
+            const requestIsStream = isStream(config.data);
             try {
-                if (isStream(config.data)) {
+                if (requestIsStream) {
                     config = { ...config, maxRedirects: 0 };
                 }
                 response = await this.axios.request(config);
@@ -155,7 +156,8 @@ export class HttpClient {
             }
 
             const apiError = new ApifyApiError(response, attempt);
-            if (this._isStatusCodeRetryable(response.status) && config.maxRedirects !== 0) {
+            if (!requestIsStream && this._isStatusCodeRetryable(response.status)) {
+                // allow a retry
                 throw apiError;
             } else {
                 stopTrying(apiError);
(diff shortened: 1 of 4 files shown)


mvolfik:
@mnmkng added the warnings as discussed

Output for stream:
```
WARN  ApifyClient: Request body was a stream - retrying will not work, as part of it was already consumed.
INFO  ApifyClient: If you want the SDK to handle retries, collect the stream into a buffer before sending it.
ERROR 
  Error: foobar
      at makeRequest (/home/volfmatej/dev/apify/actors/testing/node_modules/apify-client/dist/http_client.js:178:31)
      at RetryOperation.runAttempt [as _fn] (/home/volfmatej/dev/apify/actors/testing/node_modules/async-retry/lib/index.js:42:15)
      at RetryOperation.attempt (/home/volfmatej/dev/apify/actors/testing/node_modules/retry/lib/retry_operation.js:116:8)
      at run (/home/volfmatej/dev/apify/actors/testing/node_modules/async-retry/lib/index.js:55:8)
      at new Promise (<anonymous>)
      at retry (/home/volfmatej/dev/apify/actors/testing/node_modules/async-retry/lib/index.js:58:10)
      at HttpClient.call (/home/volfmatej/dev/apify/actors/testing/node_modules/apify-client/dist/http_client.js:148:42)
      at KeyValueStoreClient.setRecord (/home/volfmatej/dev/apify/actors/testing/node_modules/apify-client/dist/resource_clients/key_value_store.js:135:31)
      at KeyValueStore.setValue (/home/volfmatej/dev/apify/actors/testing/node_modules/apify/build/storages/key_value_store.js:230:28)
      at uploadWithSdk (/home/volfmatej/dev/apify/actors/testing/test.js:77:17)
Error: npm exited with code 91
```

Output for other values:
```
WARN  ApifyClient: API request failed 2 times. Max attempts: 5.
Cause:Error: foobar
    at makeRequest (/home/volfmatej/dev/apify/actors/testing/node_modules/apify-client/dist/http_client.js:178:31)
    at RetryOperation.runAttempt [as _fn] (/home/volfmatej/dev/apify/actors/testing/node_modules/async-retry/lib/index.js:42:15)
    at Timeout._onTimeout (/home/volfmatej/dev/apify/actors/testing/node_modules/retry/lib/retry_operation.js:85:10)
    at listOnTimeout (node:internal/timers:564:17)
    at process.processTimers (node:internal/timers:507:7)
ERROR 
  Error: foobar
      at makeRequest (/home/volfmatej/dev/apify/actors/testing/node_modules/apify-client/dist/http_client.js:178:31)
      at RetryOperation.runAttempt [as _fn] (/home/volfmatej/dev/apify/actors/testing/node_modules/async-retry/lib/index.js:42:15)
      at Timeout._onTimeout (/home/volfmatej/dev/apify/actors/testing/node_modules/retry/lib/retry_operation.js:85:10)
      at listOnTimeout (node:internal/timers:564:17)
      at process.processTimers (node:internal/timers:507:7)
Error: npm exited with code 91
```

(using a further patched version that artificially throws a retryable error)

src/http_client.ts
```diff
@@ -131,6 +133,15 @@ export class HttpClient {
         });
     }
 
+    private _informAboutStreamNoRetry() {
+        // using a module-global variable to only warn once per whole process
+        if (!warnedAboutStreamNoRetry) {
+            warnedAboutStreamNoRetry = true;
+            this.logger.warning('Request body was a stream - retrying will not work, as part of it was already consumed.');
```
mnmkng, line 140: The `logger.warning` function will automatically send the same message just once so you can utilize that :)

mvolfik: :open_mouth: TIL, thanks!

mvolfik:
Because the error messages were created for dirty local testing like this: :joy: 
```
const e = new Error("foobar");
e.config = {};
e.requst = {};
throw e;
```

is there a better way to reliably get an error response?

src/http_client.ts
```diff
@@ -130,6 +131,11 @@ export class HttpClient {
         });
     }
 
+    private _informAboutStreamNoRetry() {
+        this.logger.warning('Request body was a stream - retrying will not work, as part of it was already consumed.');
+        this.logger.info('If you want the SDK to handle retries, collect the stream into a buffer before sending it.');
```
mnmkng, line 136: The first message will be logged once, the second will be logged every time. Is this intended? I would log both as a warning.

mvolfik: ad 1] i just checked for that more carefully and turns out that even the warning is being logged repeatedly. is that a bug elsewhere? but no, the intended behavior is that both messages are logged just the first time

ad 2] oops, thanks

mnmkng: Oh, sorry. My bad. It's this one. Could you please also send a PR to the log package making an alias (to not make it a breaking change) called `warningOnce`? We've been talking about that multiple times, but seem like we've never done it.

https://github.com/apify/apify-shared-js/blob/c7978d1d93dfa53336504f3023d0be537fdf77a0/packages/log/src/log.ts#L257-L266

mvolfik: PR here: https://github.com/apify/apify-shared-js/pull/316 , will make it both a warning then. Is it better to keep it as two separate calls, so both lines are prefixed with the logging info, or should I just throw a `\n` in there?

B4nan: I would vote for renaming the function (as `warningOnce` is exactly what it does) and having `deprecated` as the alias instead.

mnmkng: @mvolfik I would just throw an `\n` there, but it's really personal preference.
@B4nan Yeah, makes sense.

mvolfik: Ooof, I just realized that the PR was released as version v2.1.0, while all deps of this package depend on 1.*. How should I proceed?

mvolfik: Pushed a fix which includes update to the dependency

mnmkng (approved):

src/http_client.ts
```diff
@@ -130,6 +131,11 @@ export class HttpClient {
         });
     }
 
+    private _informAboutStreamNoRetry() {
+        this.logger.warningOnce('Request body was a stream - retrying will not work, as part of it was already consumed.');
+        this.logger.warningOnce('If you want the SDK to handle retries, collect the stream into a buffer before sending it.');
```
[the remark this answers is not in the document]
mvolfik: Correct, sorry, lost that in the thread. Pushed a fix

drobnikj (changes requested):
Looks fine, just please add note, to why we need to extend axios config in this case.

drobnikj (approved):

PR #257, opened 2022-06-28
feat: drop single file support

valekjo:
Single file has been long deprecated and we want to get rid of it across codebase.

Theoretically this is a breaking change, but not sure if we have to treat it as such (due to long deprecation).

### DIFF
--- a/src/resource_clients/actor_version.ts
+++ b/src/resource_clients/actor_version.ts
@@ -61,13 +61,11 @@ export interface BaseActorVersion<SourceType extends ActorSourceType> {
     versionNumber?: string;
     sourceType: SourceType;
     envVars?: ActorEnvironmentVariable[];
-    baseDockerImage?: string;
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):
I don't know for sure if it's breaking (will leave that up to @B4nan) but otherwise 👍

B4nan:
You are changing publicly exported symbols, so to me this is indeed breaking, regardless of what it should be doing. Also "long deprecation" is still just a deprecation, if it was working, people might depend on it. I have no idea how breaking it actually will be, and I dont feel like me or Vlad should be the ones to decide :]

Maybe we could make the ESM switch together with this change and ship v3 to be safe?

B4nan:
v3 of client? i have no idea. v3 of crawlee will happen on wednesday

valekjo:
@B4nan - as we discussed, can we merge this now?

PR #258, opened 2022-07-04
feat: statusMessage method using the new API endpoint

barjin:
Link to API docs missing - the endpoint is not documented yet.


B4nan (approved):
dont forget to bump the version here too 

https://github.com/apify/apify-client-js/blob/master/package.json#L3

mtrunkat:
Wouldn't be better to expose the method `run(xxx).update({ statusMessage })`? To keep this more generic? It's not a big deal but at some point, we may likely add more fields.

PR #264, opened 2022-08-03
feat: Add flatten param to Dataset items listing

Strajk:
Waiting on https://github.com/apify/apify-core/pull/7537

See https://www.notion.so/apify/Dataset-fields-param-for-nested-fields-0df789303ed84faa85edf3b3cde818f0#54451f44c7944ae9980743467213d6a9

### DIFF
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -80,6 +80,7 @@ export class DatasetClient<
             clean: ow.optional.boolean,
             delimiter: ow.optional.string,
             desc: ow.optional.boolean,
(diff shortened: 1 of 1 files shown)


src/resource_clients/dataset.ts
```diff
@@ -47,6 +47,7 @@ export class DatasetClient<
         ow(options, ow.object.exactShape({
             clean: ow.optional.boolean,
             desc: ow.optional.boolean,
+            flatten: ow.optional.array.ofType(ow.string),
```
B4nan, line 50: should be added to `DatasetClientListItemOptions` too

B4nan (approved):

PR #269, opened 2022-08-19
chore: add custom catching logic for `record-or-token-not-found` errors

fnesveda:
We also want to return `undefined` when getting resources when we get a `record-or-token-not-found` error in addition to `record-not-found` errors.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -14,7 +14,6 @@ const PARSE_DATE_FIELDS_KEY_SUFFIX = 'At';
 const NOT_FOUND_STATUS_CODE = 404;
 const RECORD_NOT_FOUND_TYPE = 'record-not-found';
 const RECORD_OR_TOKEN_NOT_FOUND_TYPE = 'record-or-token-not-found';
-const NOT_FOUND_ON_S3 = '<Code>NoSuchKey</Code>';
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -38,7 +39,9 @@ export function pluckData<R>(obj: MaybeData<R>): R {
  */
 export function catchNotFoundOrThrow(err: ApifyApiError): void {
     const isNotFoundStatus = err.statusCode === NOT_FOUND_STATUS_CODE;
-    const isNotFoundMessage = err.type === NOT_FOUND_TYPE || err.message.includes(NOT_FOUND_ON_S3);
+    const isNotFoundMessage = err.type === RECORD_NOT_FOUND_TYPE
+        || err.type === RECORD_OR_TOKEN_NOT_FOUND_TYPE
+        || err.message.includes(NOT_FOUND_ON_S3);
```
mnmkng, line 44: I wonder if we still need this. I think we removed the S3 redirects some time ago, no?

mnmkng (approved):

src/utils.ts
fnesveda: You're right, we removed those ages ago. I removed the special S3 error handling too.

PR #272, opened 2022-09-05
chore(deps): "support" for TS 4.8

vladfrangu:


B4nan:
I guess we can merge this one? I see the builds to master are failing with the TS extends error... 

Not great, but for the time being, I would use `skipLibCheck` in the project templates too and call TS 4.8 generally upsupported.

PR #273, opened 2022-09-07
feat: export PaginationIterator helper class

mvolfik:


mvolfik:
wait, i realized this probably isn't really what I wanted. do we have some util for easily iterating over all values from a PaginatedList (e.g. returned by `await client.tasks().list()`?

mvolfik:
Here's what I _(well, GitHub Copilot mostly)_ came up with:
```js
import type { PaginatedList } from 'apify-client';

export async function* iterateList<T>(
    listGetter: (opts: { limit: number; offset: number }) => Promise<PaginatedList<T>>,
    limit: number,
): AsyncGenerator<T, void, never> {
    let offset = 0;
    let total = 0;
    do {
        const {
            items,
            count: returnedCount,
            total: returnedTotal,
        } = await listGetter({ limit, offset });

        offset += returnedCount;
        total = returnedTotal;
        for (const item of items) yield item;
    } while (offset < total);
}

```

Would it be worth including and exporting this util somewhere?

PR #276, opened 2022-09-22
fix: ScheduleUpdateData type

magne4000:
`ScheduleUpdateData` enforces `id` key on `actions` prop, but it should be optional.

Following snippet should be OK:
```ts
await client.schedule(schedule.id).update({
  actions: [{
    type: ScheduleActions.RunActor,
    actorId: 'someId',
  }]
});
```

But I get this error:
```
Type '{ type: ScheduleActions.RunActor; actorId: string; }' is not assignable to type 'ScheduleAction'.
  Property 'id' is missing in type '{ type: ScheduleActions.RunActor; actorId: string; }' but required in type 'ScheduleActionRunActor'.ts(2322)
schedule.d.ts(47, 5): 'id' is declared here.
```

### DIFF
--- a/src/resource_clients/schedule.ts
+++ b/src/resource_clients/schedule.ts
@@ -9,6 +9,7 @@ import {
     parseDateFields,
     catchNotFoundOrThrow,
     cast,
+    MakeOptional,
 } from '../utils';
 
 export class ScheduleClient extends ResourceClient {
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):
Thanks a lot for this, and super sorry for the fact this got lost!! 💙

PR #286, opened 2022-11-22
fix: fix isAtHome value in User-Agent header

mvolfik:


src/http_client.ts
```diff
@@ -106,7 +106,9 @@ export class HttpClient {
 
         if (isNode()) {
             // Works only in Node. Cannot be set in browser
-            const userAgent = `ApifyClient/${version} (${os.type()}; Node/${process.version}); isAtHome/${!!process.env.IS_AT_HOME}`;
+            const userAgent = `ApifyClient/${version} (${os.type()}; Node/${process.version}); isAtHome/${
+                !!process.env[ENV_VARS.IS_AT_HOME]
+            }`;
```
B4nan, line 111:
```suggestion
            const isAtHome = !!process.env[ENV_VARS.IS_AT_HOME];
            const userAgent = `ApifyClient/${version} (${os.type()}; Node/${process.version}); isAtHome/${isAtHome}`;
```

PR #289, opened 2022-12-29
feat: use common theme in v2 docs

barjin:


website/docusaurus.config.js
```diff
@@ -4,7 +4,7 @@ const { externalLinkProcessor } = require('./tools/utils/externalLink');
 const absoluteUrl = 'http://docs-v2.apify.com';
 /** @type {Partial<import('@docusaurus/types').DocusaurusConfig>} */
 module.exports = {
-    title: 'Apify Docs v2',
+    title: 'Apify JS Client',
```
B4nan, line 7: iirc this should be the same for all of the docusaurus projects, not project specific, at least thats how i originally planned it - do you use the value in the subnav or whats the reason to change it?

B4nan:
weird that the test suite is so broken, maybe something is already broken in the docs-v2 branch? but given the changes are touching only the website folder and are targetting the docs-v2, feel free to merge - the docs job passed just fine

website/docusaurus.config.js
barjin: I'm currently reusing the title in the subnav as shown in Figma, but you're right - let's keep it the same for all the projects and add another `name` field in the theme options maybe?

obrazek

B4nan: Yeah that sounds better to me.

PR #295, opened 2023-01-31
docs: unify sidebars + placeholder examples

barjin:


website/docusaurus.config.js
```diff
@@ -29,18 +29,27 @@ module.exports = {
...
+                            to: 'docs/examples',
+                            label: 'Examples',
+                            activeBaseRegex: '/docs/examples',
                         },
                         {
                             to: 'api',
-                            label: 'API',
-                            activeBaseRegex: 'client-js/api',
+                            label: 'Reference',
+                            activeBaseRegex: '/api',
```
B4nan, line 42: are you sure this is correct?

FYI we also want to change the URL to `reference` in those API links in project sites

barjin: obrazek
I'm 95% sure, does something look off to you? 

Also, the [`plugin-typedoc-api`](https://www.npmjs.com/package/docusaurus-plugin-typedoc-api) plugin we're using is specifically registering the `/api` route with no override option (that is, if we don't want to `patch-package` some more).

B4nan: > Also, the [plugin-typedoc-api](https://www.npmjs.com/package/docusaurus-plugin-typedoc-api) plugin we're using is specifically registering the /api route with no override option (that is, if we don't want to patch-package some more).

Are you sure about that? There should be `routeBasePath` option, it's just a plugin, like the docs, I believe you can control this on docusaurus level.

B4nan: right, but there is the conflict, so please rebase :]

barjin: > There should be `routeBasePath` option
yup, my bad, just got this working 👍️

PR #299, opened 2023-02-09
docs: add docs-v2

barjin:
Adds new web documentation for the project, which gets automatically deployed into the repo's github-pages env.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+    build:
+        if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+        runs-on: ubuntu-latest
+
+        steps:
+            -   uses: actions/checkout@v3
+                with:
+                    token: ${{ secrets.GH_TOKEN }}
+            -   name: Use Node.js 16
+                uses: actions/setup-node@v2-beta
```
B4nan, line 20:
```suggestion
                uses: actions/setup-node@v3
```

.gitignore
```diff
@@ -13,4 +14,15 @@ pids
...
+types
+docs/api
+docs/typedefs
+.history
+.docusaurus
+tsconfig.tsbuildinfo
+apify_storage
+crawlee_storage
+storage
+.turbo
```
B4nan, line 28: i dont think we want to pollute the root gitignore file, this should go to the website folder, no?

README.md
```diff
@@ -122,1907 +122,10 @@ const { items } = await lastSucceededRunClient.dataset().listItems();
 > will be available to all runs in the future.
 
 ### Pagination
-Most methods named `list` or `listSomething` return a [<code>Promise.&lt;PaginationList&gt;</code>](#PaginationList).
+Most methods named `list` or `listSomething` return a [<code>Promise.&lt;PaginationList&gt;</code>](/api/interface/PaginatedList).
```
B4nan, line 125: hmm i guess this wont work in the readme rendered on GH? maybe better to put absolute links in ther readme, same for the api ref

README.md
```diff
@@ -122,1907 +122,10 @@ const { items } = await lastSucceededRunClient.dataset().listItems();
...
-| --- | --- |
-| [options] | <code>object</code> |
-| [options.limit] | <code>number</code> |
-| [options.offset] | <code>number</code> |
-| [options.desc] | <code>boolean</code> |
-
-
-* * *
-
+All public classes, methods and their parameters can be inspected in the [API reference](/api).
```
B4nan, line 131: moreover this one is wrong, we changed the URL to `reference` (and it is not in the root of the website, so i guess the absolute links will be necessary)

docs/index.md
```diff
@@ -0,0 +1,22 @@
...
+---
+
+# Apify API client for JavaScript
+`apify-client` is the official library to access [Apify API](https://docs.apify.com/api/v2) from your
+JavaScript applications. It runs both in Node.js and browser and provides useful features like
+automatic retries and convenience functions that improve the experience of using the Apify API.
+
+## Quick Start
+```js
+const { ApifyClient } = require('apify-client');
```
B4nan, line 12:
```suggestion
import { ApifyClient } from 'apify-client';
```

i guess we could start using the import statements to be consistent with the rest, e.g. sdk does it already

package.json
```diff
@@ -80,16 +80,16 @@
...
 		"@types/jest": "^27.0.1",
 		"@types/node": "^16.6.1",
-		"@typescript-eslint/eslint-plugin": "^4.29.1",
-		"@typescript-eslint/parser": "^4.29.1",
-		"apify": "^2.0.2",
+		"@typescript-eslint/eslint-plugin": "^5.40.0",
+		"@typescript-eslint/parser": "^5.40.0",
 		"babel-loader": "^8.2.2",
 		"body-parser": "^1.19.0",
+		"crawlee": "^3.0.5-beta.71",
```
B4nan, line 87: uh, why is this one here? :]

website/docusaurus.config.js
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable global-require,import/no-extraneous-dependencies */
+const { config } = require('@apify/docs-theme');
+const { externalLinkProcessor } = require('./tools/utils/externalLink');
+const versions = require('./versions.json');
+
+const { absoluteUrl } = config;
+/** @type {Partial<import('@docusaurus/types').DocusaurusConfig>} */
+module.exports = {
+    title: 'Apify Docs v2',
```
B4nan, line 9:
```suggestion
    title: 'Apify Documentation',
```

this will need to be fixed in all the projects, it is used in the title tag

.gitignore
barjin: Huh, some names do not even sound like anything from docs / apify-client-js AFAIK, it looks like copy-pasted from a different repo? but agreed, moving the changes to `website` shouldn't do any harm

README.md
barjin: True, also, this README is not rendered anywhere else (docusaurus renders stuff from the docs folder only). 

What is the final URL of the new docs? `docs(-v2)?.apify.com`?

B4nan: > True, also, this README is not rendered anywhere else (docusaurus renders stuff from the docs folder only).

Oh, I thought we render it in the API reference as in the SDK, but I see it is not there (which is fine, as it would be duplicated).

> What is the final URL of the new docs? docs(-v2)?.apify.com?

https://docs.apify.com, and the sdk and developers subdomains will be redirected to it

barjin:
It's a bit weird that the tests aren't passing - it seems that some upstream dependencies cannot [get resolved in the jest tests](https://github.com/apify/apify-client-js/actions/runs/4133445826/jobs/7143359470#step:6:271). Any ideas why this might be happening?

B4nan:
Maybe let's revert the changes in the root package.json/lock file? This PR should be mostly about adding the docs/website folders.

B4nan:
Or better do them first in master to see if they break it, we want to have up to date deps but mixing it in this PR can only cause confusing.

B4nan:
I think [this commit](https://github.com/apify/apify-client-js/pull/299/commits/737fd97ba2036f3c667e95bc3afa02df50605762) sneaked in, I got it in master unpushed as it breaks the tests, and I mistakenly added it to the docs-v2 branch when starting it. It's probably failing as we compile via webpack and it does not understand the external dependencies.

So I guess we can just revert it from the docs-v2 and get back to the update later.

PR #330, opened 2023-03-07
feat: add support for `maxItems` in run options

novotnyj:
This adds support for optional `maxItems` param to `start` and `call` methods on actors and tasks. 

This param will be used by pay per result actors to limit max items that will be charged to customer.

Should be merged after https://github.com/apify/apify-worker/pull/734 is released.

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -329,7 +329,7 @@ export interface ActorStartOptions {
     webhooks?: readonly WebhookUpdateData[];
 
     /**
-     * Specifies maximum number of items that the actor run should process.
+     * Specifies maximum number of items that the actor run should return.
      * This is used by pay per result actors to limit the maximum number of results that will be charged to customer.
      * Value can be accessed in actor run using `APIFY_ACTOR_MAX_ITEMS` environment variable.
      */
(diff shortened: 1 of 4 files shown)


src/resource_clients/actor.ts
```diff
@@ -324,6 +327,13 @@ export interface ActorStartOptions {
      * [ad hook webhooks documentation](https://docs.apify.com/webhooks/ad-hoc-webhooks) for detailed description.
      */
     webhooks?: readonly WebhookUpdateData[];
+
+    /**
+     * Specifies maximum number of items that the actor run should process.
```
mhamas, line 332: should "return", not "process"

test/actors.test.js
```diff
@@ -208,6 +220,7 @@ describe('Actor methods', () => {
             const data = { id: runId, actId: actorId, status: 'SUCCEEDED' };
             const body = { data };
             const waitSecs = 1;
+            const maxItems = 100;
```
mhamas, line 223: is there a unit test that also works without `maxItems`? To verify that this param is really optional.

mhamas (changes requested):

mhamas (approved):

vladfrangu (approved):

B4nan:
> Should be merged after https://github.com/apify/apify-worker/pull/734 is released.

I hope it won't mind I merged this @novotnyj, will hold the stable release before that PR is merged.

novotnyj:
It's ok, it won't break anything🙂

PR #331, opened 2023-03-07
fix(actor): fix types for actor run

drobnikj:
see https://github.com/apify/apify-core/issues/8077 for context

Plus I found some missing attrs, which we have in API but not in types and not in API ref.

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -330,25 +330,46 @@ export interface ActorCallOptions extends Omit<ActorStartOptions, 'waitForFinish
     waitSecs?: number;
 }
 
-export interface ActorRun {
+export interface ActorRunList {
     id: string;
     actId: string;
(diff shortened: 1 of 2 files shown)


src/resource_clients/actor.ts
```diff
@@ -330,25 +330,46 @@ export interface ActorCallOptions extends Omit<ActorStartOptions, 'waitForFinish
     waitSecs?: number;
 }
 
-export interface ActorRun {
+export interface ActorRunList {
```
fnesveda, line 333: This name suggests that this interface describes the list of actor runs, but this is still just a single run.

How about calling this `ActorRunListItem`?

fnesveda (changes requested):
One naming suggestion, otherwise looks good 👍

fnesveda (approved):

PR #335, opened 2023-03-10
feat(repo): repository links

drobnikj:


drobnikj:
@B4nan I'm not sure, but I would expect a redirect in this case. Maybe worth checking if we have somewhere https://apify.com/docs/sdk/apify-client-js/latest

B4nan:
There is apparently one redirect already for the https://apify.com/docs/sdk/apify-client-js/latest -> https://docs.apify.com/sdk/apify-client-js/latest, so ideally that one should be fixed instead of adding another to the chain. Any idea where that might be implemented?

B4nan (approved):

drobnikj:
@B4nan not sure but just add
/sdk/apify-client-js/latest -> /api/client/js/ should work

B4nan:
Yeah already fixing this.

PR #373, opened 2023-07-12
feat: use Actor/Apify env vars instead of `ENV_VARS`

jirimoravcik:


src/apify_client.ts
```diff
@@ -312,8 +312,9 @@ export class ApifyClient {
     }
 
     async setStatusMessage(message: string, options?: SetStatusMessageOptions): Promise<void> {
-        const runId = process.env[ENV_VARS.ACTOR_RUN_ID];
-        if (!runId) { throw new Error(`Environment variable ${ENV_VARS.ACTOR_RUN_ID} is not set!`); }
+        // TODO: Keep backwards compatibility or just do a breaking change?
```
drobnikj, line 315: I think we can make it breaking as it is used only on the platform, where we already fix it.

fnesveda (approved):
Looks good, just a note that we need to deploy the worker changes before we release this

jirimoravcik:
> Looks good, just a note that we need to deploy the worker changes before we release this

Yeah, you're right. I forgot to mention it in the PR, sorry.

PR #379, opened 2023-07-21
Update index.js

jancurn:


B4nan (approved):
I will be fixing the `<title>` tags in all the repo today, so will first give you a screenshot of how it looks now. Or we can merge now and see it in the wild (and improve on that) if you want, feel free to merge.

jancurn:
cheers @B4nan I think it's okay to merge, we can always change it later

PR #395, opened 2023-08-24
feat: Added StoreCollectionClient class useful for listing Actors in Apify Store

drobnikj:
New method for listing Actors in store
See API reference https://docs.apify.com/api/v2/#/reference/store

### DIFF
--- a/src/resource_clients/store_collection.ts
+++ b/src/resource_clients/store_collection.ts
@@ -1,0 +1,69 @@
+import ow from 'ow';
+import { ApiClientSubResourceOptions } from '../base/api_client';
+import { ResourceCollectionClient } from '../base/resource_collection_client';
+import { PaginatedList } from '../utils';
+import { ActorStats } from './actor';
+
+/**
+ * @hideconstructor
+ */
+export class StoreCollectionClient extends ResourceCollectionClient {
+    constructor(options: ApiClientSubResourceOptions) {
+        super({
+            resourcePath: 'store',
+            ...options,
+        });
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


fnesveda (approved):
Looks good, and works well!

src/resource_clients/store_collection.ts
```diff
@@ -0,0 +1,69 @@
+import ow from 'ow';
+import { ApiClientSubResourceOptions } from '../base/api_client';
+import { ResourceCollectionClient } from '../base/resource_collection_client';
+import { PaginatedList } from '../utils';
+import { ActorStats } from './actor';
+
+/**
+ * @hideconstructor
```
B4nan, line 8: FYI this has no effect on the generated docs, you need to put a `@hidden` jsdoc over the constructor itself (this problem is present in many other places in this repo)

src/resource_clients/store_collection.ts
```diff
@@ -0,0 +1,69 @@
...
+     * https://docs.apify.com/api/v2/#/reference/store/store-actors-collection/get-list-of-actors-in-store
+     * @param {object} [options]
+     * @param {number} [options.limit]
+     * @param {number} [options.offset]
+     * @param {string} [options.search]
+     * @param {string} [options.sortBy]
+     * @param {string} [options.category]
+     * @param {string} [options.username]
+     * @param {string} [options.pricingModel]
+     * @return {Promise<PaginationList>}
```
B4nan, line 28: i'd remove this, it adds no value, the information is all available on type level (and it wont have any effect on the docs either)

B4nan (approved):
just two nits

Issue #399, opened 2023-08-31
Does not work on TS project? agentkeepalive import


DavideViolante:
I have the following code in a TS project:

### Code
```ts
const client = new ApifyClient({
  token: process.env.APIFY_TOKEN,
})
;

...

  const run = await client.actor('shanes/tweet-flash').call(input);
  const { items } = await client.dataset(run.defaultDatasetId).listItems();

...   
```

But the code does not compile, returning this error:

### Error
```
> tsc -p server

node_modules/apify-client/dist/http_client.d.ts:2:8 - error TS1259: Module '"/home/dave/job/utopia/utopia/node_modules/agentkeepalive/index"' can only be default-imported using the 'esModuleInterop' flag

2 import KeepAliveAgent from 'agentkeepalive';
         ~~~~~~~~~~~~~~

  node_modules/agentkeepalive/index.d.ts:62:1
    62 export = AgentKeepAlive;
       ~~~~~~~~~~~~~~~~~~~~~~~~
    This module is declared with 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.


Found 1 error in node_modules/apify-client/dist/http_client.d.ts:2
```

Note: changing esModuleInterop flag is not a proper fix.

### Possible fix file: apify-client/dist/http_client.d.ts
```ts
import KeepAliveAgent from 'agentkeepalive';
```

```ts
import * as KeepAliveAgent from 'agentkeepalive';
```

DavideViolante:
Repro:
```ts
import { ApifyClient } from 'apify-client';

const client = new ApifyClient({
  token: 'YOUR_TOKEN',
});

async function main(): Promise<any> {
  try {
    const run = await client.actor('shanes/tweet-flash').call({});
    const { items } = await client.dataset(run.defaultDatasetId).listItems();
    console.log(items)
  } catch (err) {
    console.log(err);
  }
}

main();

```

```json
{
  "name": "tsapify",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "build": "tsc",
    "start": "node index",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@types/node": "^20.5.9",
    "typescript": "^5.2.2"
  },
  "dependencies": {
    "apify-client": "^2.7.2"
  }
}

```
```json
{
  "compilerOptions": {
[...]

PR #400, opened 2023-09-01
Fix KeepAliveAgent import

DavideViolante:
Fix https://github.com/apify/apify-client-js/issues/399

Tested on my project with this version and it seems to work fine. Needs further testing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "apify-client",
-	"version": "2.7.2",
+	"version": "2.7.3",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "apify-client",
-			"version": "2.7.2",
+			"version": "2.7.3",
 			"license": "Apache-2.0",
 			"dependencies": {
 				"@apify/consts": "^2.19.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "apify-client",
-	"version": "2.7.2",
+	"version": "2.7.3",
 	"description": "Apify API client for JavaScript",
 	"main": "dist/index.js",
 	"module": "dist/index.mjs",
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -1,7 +1,7 @@
 import os from 'os';
 import axios, { AxiosError, AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import retry, { RetryFunction } from 'async-retry';
-import KeepAliveAgent from 'agentkeepalive';
+import KeepAliveAgent = require('agentkeepalive');
 import { APIFY_ENV_VARS } from '@apify/consts';
 import { Log } from '@apify/log';
 import { ApifyApiError } from './apify_api_error';
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -1,7 +1,7 @@
 import os from 'os';
 import axios, { AxiosError, AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import retry, { RetryFunction } from 'async-retry';
-import * as KeepAliveAgent from 'agentkeepalive';
+import KeepAliveAgent = require('agentkeepalive');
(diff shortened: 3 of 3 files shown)


B4nan:
This is not really a fix, more like a workaround, as `require` is weakly typed. I would rather slam a `@ts-ignore` comment over this line, it should help the same way. 

That error message sounds like doing this should help too:

```ts
import * as KeepAliveAgent from 'agentkeepalive';
```

But that will probably require some code adjustments, I will take a closer look later today.

DavideViolante:
You are right @B4nan . During my first attempt I also used `import *` but didn't work (idk exactly why).
Now I tried it again on my project and it seems to work fine.
I sent a commit with the new import.

DavideViolante:
> which means that all the projects using our libraries should be using it as well

This is wrong because projects using this library _should not be forced_ to set that flag to true or false.
I think the goal of any library is to work on both these scenarios (and many do).  Both scenarios are CommonJS and ESM (not just the esModuleInterop flag true/false). 

I recommend to read these:
- https://evertpot.com/universal-commonjs-esm-typescript-packages/
- https://transang.me/how-to-publish-a-hybrid-npm-package/

DavideViolante:
In the past TS was generating tsconfig with that flag off, as far as I remember. I'm working with a pretty old and big project so that flag was never changed, since it never raised any problem (until now :)).

B4nan:
And what if you enable it, do you get some new problems? I don't think you should.

---

I am not opposed to disable the flag, especially not in the apify client (not so much in the rest of our node only projects, because of what I said above with the `module: node16` inference). But since you are the very first person complaining about it (and we have this flag enabled in all of our TS projects, for about 2 years now), I really don't think it's problematic to just enable it. IIRC it's a problem to disable it (as that forces you to use star imports instead of default imports), but not the other way.

DavideViolante:
I think you misunderstood, you should not disable the flag in this project, but I think the generated build should work _regardless_ of the flag value. Generally, I think it's wrong to force the final user to change their configuration.
Probably my solution is not the best one, because it should be changed how the build is generated. This issue will happen again for any package like agentkeepalive.

But in the end it's a _choice_ to not support "old" projects and I totally understand.

B4nan:
And I think you misunderstood me. The point of using `esModuleInterop` is to be able to use the default imports (as we do now), and it really means the downstream projects need to have it enabled too, because of how it affects the transpilation. So saying we should keep the flag enabled but change our code back makes very little sense to me. This is exactly the way we wouldn't be warning about the problem on compile time...

And once again, I don't mind disabling the flag here (so it does not affect downstream, and we are warned about this problem on compile time).

> Yes, I receive several issues after changing all my imports. 

I said you should try to enable the flag, not to change anything in your code. I don't think you need to change the star imports, the flag is about default imports acting the same way as star imports, not about forcing you to change star imports to default imports.

DavideViolante:
> I said you should try to enable the flag, not to change anything in your code. I don't think you need to change the star imports, the flag is about default imports acting the same way as star imports, not about forcing you to change star imports to default imports.

This is not correct because as soon as I enable the esModuleInterop flag, I can't keep the same code because, at compile type, I receive all the errors about the imports, like this one (eg: express, dotenv, morgan, cors, supertest, etc):
```ts
server/app.ts:10:13 - error TS2349: This expression is not callable.
  Type 'typeof e' has no call signatures.

10 const app = express();
               ~~~~~~~

  server/app.ts:3:1
    3 import * as express from 'express';
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```
You can easly try on [this project](https://github.com/DavideViolante/Angular-Full-Stack).
Only changing all my imports I can compile correctly. (it's very easy to do on small projects, I know)
But, on a bigger project I'm working on, some libraries are also returning errors at runtime, for example those that must be imported as destructured.

B4nan:
And this code worked for you without the flag? The code seems wrong, if you use star import, it does not represent any function, but a dictionary of the exported symbols.

PR #412, opened 2023-09-11
fix: don't parse non-date strings

### LINKED ISSUE #406: Improve parsing of dates in user data
Currently, we try to parse all response fields ending with `At` as dates, even in user-supplied data in dataset items, requests in queues etc. This leads to undesired behavior when the data is not parseable as dates.

For example, this request
```jsonc
{
    "url": "http://example.com",
    "userData": {
        "postedAt": "30+ days ago",
    },
}
```
would get parsed as 
```js
{
    url: "http://example.com",
    userData: {
        postedAt: Invalid Date,
    },
}
```
and there would be no way to get to the actual `postedAt` field contents.

We could either:
- **only parse the date fields if they are actual dates, and keep the original values if they're not (easy, not very breaking)**
- not parse dates in user-supplied data at all, parse just data coming from our own data structures (harder, breaking, but feels more proper)

barjin:
Non-date strings in `fieldAt` fields of `userData` are being converted to `Date` nonetheless, leading to replacing the actual values with `Invalid Date`. 

This PR first checks whether the `Date` conversion resulted in an actual date and only then stores the date - otherwise, the original value is preserved.

Fixes #406

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -60,7 +60,7 @@ export function parseDateFields(input: JsonValue, depth = 0): ReturnJsonValue {
         if (k.endsWith(PARSE_DATE_FIELDS_KEY_SUFFIX)) {
             if (v) {
                 const d = new Date(v as string);
-                output[k] = isNaN(d.getTime()) ? v as string : d;
+                output[k] = Number.isNaN(d.getTime()) ? v as string : d;
             } else {
                 output[k] = v;
             }


src/utils.ts
```diff
@@ -58,7 +58,13 @@ export function parseDateFields(input: JsonValue, depth = 0): ReturnJsonValue {
     return Object.entries(input).reduce((output, [k, v]) => {
         const isValObject = !!v && typeof v === 'object';
         if (k.endsWith(PARSE_DATE_FIELDS_KEY_SUFFIX)) {
-            output[k] = v ? new Date(v as string) : v;
+            if (v) {
+                const d = new Date(v as string);
+                // eslint-disable-next-line no-self-compare
+                output[k] = d.getTime() === d.getTime() ? d : v as string;
```
B4nan, line 64: maybe better this way?

```suggestion
                output[k] = isNaN(d.getTime()) ? v as string : d;
```

B4nan (approved):

B4nan:
ha, PR toolkit finally works :]

PR #416, opened 2023-09-18
chore: Removed references to issuesEnabled

Jkuzz:
part of https://github.com/apify/apify-core/issues/10703

### DIFF
--- a/website/versioned_docs/version-2.6/api-typedoc.json
+++ b/website/versioned_docs/version-2.6/api-typedoc.json
@@ -1278,6 +1278,10 @@
 															},
 															{
 																"type": "literal",
+																"value": "issuesEnabled"
+															},
+															{
+																"type": "literal",
 																"value": "title"
 															},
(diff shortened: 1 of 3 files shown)


HonzaTuron (approved):

website/versioned_docs/version-2.6/api-typedoc.json
```diff
@@ -1276,10 +1276,6 @@
 																"type": "literal",
```
mhamas, line 1276: I believe these `versioned_docs` are autogenerated so no need to change it here. Also, you are changing the history (this is version 2.6. and the other file is 2.7 while the current version is 2.8) which is not correct.

mhamas (approved):
Just please remove the changes from versioned docs, pre-approving.

B4nan:
yes, please remove the changes from versioned docs. i will create new version snapshot for 2.8 once this is merged.

PR #440, opened 2023-10-25
feat: Add how to install javascript Apify client

webrdaniel:


B4nan (changes requested):
you need to update the next version too, this is just a version snapshot

https://github.com/apify/apify-client-js/blob/master/docs/index.md

i would also generally discourage the prose wrap (line breaks in markdown), but that's something that should be enforced via linter on the whole docs (we do it in the main docs repo already)

jbartadev (approved):
Cool.

@B4nan wouldn't be enough to just update the `index.md` file? Do you also need to always update the specific version file?

B4nan:
depends on what you want, but generally nope:
- current stable version is the last versioned one, so https://docs.apify.com/api/client/js/docs is `website/versioned_docs/version-2.8/index.md`
- next version (will be used for creating next stable version snapshot) is at https://docs.apify.com/api/client/js/docs/next `docs/index.md`

so when you update only the next version, its fine, but you wont see any changes in the stable docs (untill we ship new minor version). when you update only the snapshot, your changes will be lost on next minor release (they will only stay in that specific snapshot)

B4nan (approved):

webrdaniel:
> depends on what you want, but generally nope:
> 
> * current stable version is the last versioned one, so https://docs.apify.com/api/client/js/docs is `website/versioned_docs/version-2.8/index.md`
> * next version (will be used for creating next stable version snapshot) is at https://docs.apify.com/api/client/js/docs/next `docs/index.md`
> 
> so when you update only the next version, its fine, but you wont see any changes in the stable docs (untill we ship new minor version). when you update only the snapshot, your changes will be lost on next minor release (they will only stay in that specific snapshot)

@B4nan Thanks for clarification 😄

PR #451, opened 2023-11-09
fix(schedule): expose other fields when id optional

omikader:
This PR appears to follow-up on https://github.com/apify/apify-client-js/pull/276 which made it so `id` was optional ✅ but did so in a way that removed the remainder of the required fields from the type definition ❌, e.g. `actorTaskId` was no longer recognized as a valid property as of v2.8.2

Because `ScheduleAction` is defined as a union of different types, we must use [distributive conditional types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) to capture the specific properties of the individual subtypes correctly.

### DIFF
--- a/src/resource_clients/schedule.ts
+++ b/src/resource_clients/schedule.ts
@@ -9,7 +9,7 @@ import {
     parseDateFields,
     catchNotFoundOrThrow,
     cast,
-    MakeOptional,
+    DistributiveOmit,
 } from '../utils';
 
 export class ScheduleClient extends ResourceClient {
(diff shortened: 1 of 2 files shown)


B4nan (approved):

omikader:
Hi @B4nan -- thanks for reviewing! Would it be possible to get an official release of [`apify-client`](https://www.npmjs.com/package/apify-client) with this fix? Looks like the last stable version is `2.8.2` from 3 weeks ago?

Thanks again!

PR #246, opened 2022-05-18
feat: add request queue methods `listAndLockHead`, `prolongRequestLock`, `deleteRequestLock`

drobnikj:
Add methods:

* listAndLockHead
* prolongRequestLock
* deleteRequestLock

PR #468, opened 2023-12-06
feat: add `runs()` and `builds()` top level endpoints

### LINKED ISSUE #296: Add `runs()` and `builds()` top level endpoints
They're commented out and not implmented, but we already have both of them.

https://docs.apify.com/api/v2#/reference/actor-builds/build-collection/get-user-builds-list
https://docs.apify.com/api/v2#/reference/actor-runs/run-collection/get-user-runs-list

https://github.com/apify/apify-client-js/blob/80008ea77bb16211cc177cae377c8e220cff3ed1/src/apify_client.ts#L109-L115

foxt451:
Closes #296

### DIFF
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -29,6 +29,8 @@ import { WebhookCollectionClient } from './resource_clients/webhook_collection';
 import { WebhookDispatchClient } from './resource_clients/webhook_dispatch';
 import { WebhookDispatchCollectionClient } from './resource_clients/webhook_dispatch_collection';
 import { Statistics } from './statistics';
+import { BuildCollectionClient } from './resource_clients/build_collection';
+import { RunCollectionClient } from './resource_clients/run_collection';
 
 /**
  * ApifyClient is the official library to access [Apify API](https://docs.apify.com/api/v2) from your
@@ -107,13 +109,9 @@ export class ApifyClient {
         });
     }
 
-    // TODO we don't have this endpoint yet
-    // /**
-    //  * @return {BuildCollectionClient}
-    //  */
-    // builds() {
-    //     return new BuildCollectionClient(this._options());
-    // }
+    builds(): BuildCollectionClient {
+        return new BuildCollectionClient(this._options());
+    }
 
     /**
      * https://docs.apify.com/api/v2#/reference/actor-builds/build-object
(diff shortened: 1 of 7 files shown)


B4nan:
> @[drobnikj](https://github.com/drobnikj) drobnikj (Jakub Drobník) requested review from [B4nan](https://github.com/B4nan) and [drobnikj](https://github.com/drobnikj) 

This is not for our team to evaluate, I don't know the story behind it, nor the state of our API when it comes to supporting this.

src/apify_client.ts
```diff
@@ -108,13 +110,9 @@ export class ApifyClient {
...
     }
 
-    // TODO we don't have this endpoint yet
-    // /**
-    //  * @return {BuildCollectionClient}
-    //  */
-    // builds() {
-    //     return new BuildCollectionClient(this._options());
-    // }
+    builds(): BuildCollectionClient {
```
drobnikj, line 113: Please add a reference to API docs into comments, the same as we had in the rest of the endpoints.

src/apify_client.ts
```diff
@@ -204,13 +202,12 @@ export class ApifyClient {
...
     }
 
-    // TODO we don't have this endpoint yet
-    // /**
-    //  * @return {RunCollectionClient}
-    //  */
-    // runs() {
-    //     return new RunCollectionClient(this._options());
-    // }
+    runs(): RunCollectionClient {
```
drobnikj, line 205: Same as the build list, please add a reference to docs.

drobnikj (changes requested):
Thanks for pushing this forward. 
It looks good, can you please add unit tests for new methods?
You can check [here](https://github.com/apify/apify-client-js/pull/246/files#diff-ff1a96a183494b9c296289da90a1171bbc84c8e707c9e5ba5eda2aeac6b0b1fb) for example to have idea how to do it.

drobnikj (approved):
👍

B4nan:
FYI I did some dependency updates including the shared eslint config, which now enforces import order, that's the reason for the conflicts. Those problems are autofixable.

B4nan:
I'd like to ship new minor version, would be good to include this there, @foxt451 can you please rebase and fix the linting issues?

PR #506, opened 2024-02-06
fix: publish browser bundle

B4nan:
For some unknown reason, we stopped publishing it in v2.0.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,12 +5,15 @@
 	"main": "dist/index.js",
 	"module": "dist/index.mjs",
 	"types": "dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,12 +5,15 @@
 	"main": "dist/index.js",
 	"module": "dist/index.mjs",
 	"types": "dist/index.d.ts",
+    "browser": "dist/bundle.js",
```
vladfrangu, line 8: The intents seem wrong but maybe its just mobile github being silly

vladfrangu (approved):
Looks good, worlds smallest non blocking nit

PR #510, opened 2024-02-12
feat: add `KeyValueStore.recordExists()` method

### LINKED ISSUE #507: KeyValueStore: Method to check key-value store record exists
It is impossible to check if a key-value record exists, and users use `kvs.getRecord` instead, which is inefficient.

The implementation is simple: to check if a record exists, you can use an HTTP request with the method HEAD get Record API URL.

Method:
`keyValueStoreClient.exists(key: string): boolean`
which returns true or false based on the record.

barjin:
Closes #507

### DIFF
--- a/src/resource_clients/key_value_store.ts
+++ b/src/resource_clients/key_value_store.ts
@@ -70,6 +70,23 @@ export class KeyValueStoreClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    async recordExists(key: string): Promise<boolean> {
+        const requestOpts: Record<string, unknown> = {
+            url: this._url(`records/${key}`),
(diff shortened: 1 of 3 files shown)


src/resource_clients/key_value_store.ts
```diff
@@ -70,6 +70,23 @@ export class KeyValueStoreClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    async recordExists(key: string): Promise<boolean> {
```
B4nan, line 73: its a public method, lets add at least a basic JSDoc explaining what it does

B4nan (approved):

drobnikj (approved):
Nice, that was fast!

PR #517, opened 2024-02-21
feat: add monthlyUsage() and limits() endpoints to UserClients

tobice:
These endpoints have been around for some quite time and now we're exposing them via the JS client.

### DIFF
--- a/src/resource_clients/user.ts
+++ b/src/resource_clients/user.ts
@@ -1,5 +1,8 @@
+import { ApifyApiError } from '../apify_api_error';
 import { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceClient } from '../base/resource_client';
+import { ApifyRequestConfig } from '../http_client';
+import { cast, catchNotFoundOrThrow, parseDateFields, pluckData } from '../utils';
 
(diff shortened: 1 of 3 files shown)


tobice:
The data responses are quite big and so are the TypeScript interfaces. I honestly can't say they are 100% correct. I used ChatGPT to transform a sample API response into an interface and then I cross-checked it with the code base (e.g. all fields seem to be mandatory). But I might have missed something.

Once this is released, I'll extend integration tests in `apify-core` to make sure that all fields are populated.

jirimoravcik (approved):

jirimoravcik:
> The data responses are quite big and so are the TypeScript interfaces. I honestly can't say they are 100% correct. I used ChatGPT to transform a sample API response into an interface and then I cross-checked it with the code base (e.g. all fields seem to be mandatory). But I might have missed something.
> 
> Once this is released, I'll extend integration tests in `apify-core` to make sure that all fields are populated.

Yeah, in the future, we want to have types in API and share them across our codebase, sadly, we're not there yet.

drobnikj (approved):

Issue #183, opened 2021-07-26
Unable to resolve module events from node_modules/@apify/log/logger.js:


karthikeyan-kandasamy:
Getting this error Unable to resolve module events from /Users/hubino/Documents/RN_Projects/Homi/node_modules/@apify/log/logger.js:
Is this package supports Expo React native?

mnmkng:
Hmm, interesting. `events` is a Node.js only module, so it makes sense that it's not available in React Native. Interestingly, we've been using `apify-client` in the browser for quite some time without any problems. I guess Webpack polyfills it for us.

Can you polyfill `events` in Expo?

vdusek:
Closing as a duplicate of #753

PR #519, opened 2024-02-22
feat: parse monthlyUsage.dailyServiceUsages[].date as Date

tobice:
This field does not end with "At", which means it is ignored by the existing parseDateFields helper. This PR adjusts this helper to support parsing other fields as well.

To avoid accidental braking changes by applying the new parser on all existing data structures, the helper now has a new optional shouldParseField() param to identify other fields to be parsed. Without this param, the helper behaves just like before.

The actual user-facing change (`typeof monthlyUsage.dailyServiceUsages[].date === Date`) was tested manually and will be covered by the integration tests in `apify-core`.

### DIFF
--- a/src/resource_clients/user.ts
+++ b/src/resource_clients/user.ts
@@ -35,7 +35,10 @@ export class UserClient extends ResourceClient {
         };
         try {
             const response = await this.httpClient.call(requestOpts);
-            return cast(parseDateFields(pluckData(response.data)));
+            return cast(parseDateFields(
+                pluckData(response.data),
+                // Convert  monthlyUsage.dailyServiceUsages[].date to Date (by default it's ignored by parseDateFields)
+                /* shouldParseField = */ (key) => key === 'date'));
         } catch (err) {
             catchNotFoundOrThrow(err as ApifyApiError);
         }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,6 +2,7 @@ import type { Readable } from 'node:stream';
 import util from 'util';
 import zlib from 'zlib';
 
+import log from '@apify/log';
 import ow from 'ow';
 import type { TypedArray, JsonValue } from 'type-fest';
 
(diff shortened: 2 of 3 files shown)


src/resource_clients/user.ts
```diff
@@ -35,7 +35,10 @@ export class UserClient extends ResourceClient {
         };
         try {
             const response = await this.httpClient.call(requestOpts);
-            return cast(parseDateFields(pluckData(response.data)));
+            return cast(parseDateFields(
+                pluckData(response.data),
+                // Convert  monthlyUsage.dailyServiceUsages[].date to Date (by default it's ignored by parseDateFields)
+                /* shouldParseField = */ (key) => key === 'date'));
```
tobice, line 41: @B4nan FYI this is technically speaking a breaking change after #183 as it changes `string` to `Date`. But #183 hasn't been released yet, so it should be fine. We just need to wait with the client release.

src/utils.ts
```diff
@@ -51,24 +49,34 @@ type ReturnJsonObject = { [Key in string]?: ReturnJsonValue; };
...
+ * If you want parse other fields as well, you can provide a custom matcher function shouldParseField(). This
+ * admittedly awkward approach allows this function to be reused for various purposes without introducing potential
+ * breaking changes.
+ *
+ * If the field cannot be converted to Date, it is left as is.
  */
-export function parseDateFields(input: JsonValue, depth = 0): ReturnJsonValue {
+export function parseDateFields(input: JsonValue, shouldParseField: ((key: string) => boolean) | null = null, depth = 0): ReturnJsonValue {
+    const PARSE_DATE_FIELDS_KEY_SUFFIX = 'At';
+    const PARSE_DATE_FIELDS_MAX_DEPTH = 3; // obj.data.someArrayField.[x].field
```
jirimoravcik, line 63: (nit) I think we rarely put const definitions inside the function like this. Generally, the consts are named in a way that makes it clear for what they're meant and it's easier to reason about them when they're all defined together at the top of the file.

jirimoravcik (approved):

src/utils.ts
tobice: Here is my reasoning:

* I find it useful for the method's documentation (or implementation for that matter) to be straightforward about what it does
* That means explicitly mentioning the "At" suffix
* In order to keep the code in sync (should that ever change), all the mentions should be as close to each other possible

Moreover:
* Both constants are really only used here
* It is not configuration in the true sense

So, I decided to just inline the values 😅

Alternatives:
* Rename to `parseDateFieldsKeySuffix` (snake case should be less "triggering")
* Move `parseDateFields` to a standalone file. Then the constants can stay as they are as they will be clearly scoped.

src/utils.ts
```diff
@@ -51,24 +49,34 @@ type ReturnJsonObject = { [Key in string]?: ReturnJsonValue; };
 type ReturnJsonArray = Array<ReturnJsonValue>;
 
 /**
- * Helper function that traverses JSON structure and parses fields such as modifiedAt or createdAt to dates.
+ * Traverses JSON structure and converts fields that end with "At" to a Date object (fields such as "modifiedAt" or
+ * "createdAt").
+ *
+ * If you want parse other fields as well, you can provide a custom matcher function shouldParseField(). This
+ * admittedly awkward approach allows this function to be reused for various purposes without introducing potential
```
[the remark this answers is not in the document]
tobice: It is what it is 🤷😅

These conversions should be either explicit (as in, manually convert the whole response field by field) or automated using an actual contract, not just naming convention (OpenAPI). Let's hope we can get there 🤞

tobice:
I can see a 👍 from B4nan and an approval from Jirka, so resolving and merging to get unblocked 😊 

Feel free to leave more feedback and I'll fix it in follow-up PRs.

src/utils.ts
jirimoravcik: inlining is ok

src/utils.ts
```diff
@@ -51,24 +50,37 @@ type ReturnJsonObject = { [Key in string]?: ReturnJsonValue; };
...
+        return input as ReturnJsonValue;
+    }
+
+    if (Array.isArray(input)) return input.map((child) => parseDateFields(child, shouldParseField, depth + 1));
     if (!input || typeof input !== 'object') return input;
 
     return Object.entries(input).reduce((output, [k, v]) => {
         const isValObject = !!v && typeof v === 'object';
-        if (k.endsWith(PARSE_DATE_FIELDS_KEY_SUFFIX)) {
+        if (k.endsWith('At') || (shouldParseField && shouldParseField(k))) {
```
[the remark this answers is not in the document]
tobice: Ah, didn't know that, thanks 👍

But given how minor this is I won't be sending a dedicated PR but I will remember it for the next time.

PR #521, opened 2024-02-26
chore: remove warning when parseDateFields reaches depth limit

tobice:
This is polluting logs as there are data structures in API responses that have arbitrary depth (user data, data sets).

Context:
https://apifier.slack.com/archives/CGZSN9DQC/p1708707905914559


B4nan (approved):

tobice:
Issue for a proper solution: https://github.com/apify/apify-client-js/issues/522

PR #523, opened 2024-02-29
fix: add missing `isApifyIntegration` field to `Webhook` type

omikader:
This PR adds a missing field on the `Webhook` interface. As per the [documentation](https://docs.apify.com/platform/integrations/actors/integrating-actors-via-api), one can programmatically integrate actors together using the "create webhook" endpoint. The `isApifyIntegration` field helps the UI distinguish whether to render it as an ordinary HTTP webhook in the Apify console.

### DIFF
--- a/src/resource_clients/webhook.ts
+++ b/src/resource_clients/webhook.ts
@@ -93,6 +93,7 @@ export interface Webhook {
     lastDispatch: string;
     stats: WebhookStats;
     shouldInterpolateStrings: boolean;
+    isApifyIntegration: boolean;
     headersTemplate?: string;
     description?: string;
 }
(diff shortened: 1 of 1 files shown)


src/resource_clients/webhook.ts
```diff
@@ -93,6 +93,7 @@ export interface Webhook {
     lastDispatch: string;
     stats: WebhookStats;
     shouldInterpolateStrings: boolean;
+    isApifyIntegration: boolean;
```
valekjo, line 96:
```suggestion
    isApifyIntegration?: boolean;
```

valekjo (commented):
The field should be optional, otherwise it looks good, thanks!

src/resource_clients/webhook.ts
omikader: Thanks for the quick review @valekjo! I assumed it should be optional in `WebhookUpdateData` via `Partial` but required in `Webhook` as it is always part of the response.

All good either way 😄

PR #559, opened 2024-06-18
feat: added data property to API error object

gippy:
This update is needed because in API we will be adding a new data field to the API error, which will be used during dataset validation.

https://github.com/apify/apify-docs/pull/1079

### DIFF
--- a/test/apify_api_error.test.js
+++ b/test/apify_api_error.test.js
@@ -4,13 +4,16 @@ const { ApifyClient } = require('../src/index');
 
 describe('ApifyApiError', () => {
     const browser = new Browser();
+    let server;
 
     beforeAll(async () => {
         await browser.start();
+        server = await mockServer.start();
     });
 
     afterAll(async () => {
         await browser.cleanUpBrowser();
+        await mockServer.close();
     });
 
     test('should carry all the information', async () => {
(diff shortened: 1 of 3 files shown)


test/apify_api_error.test.js
```diff
@@ -62,4 +63,56 @@ describe('ApifyApiError', () => {
         expect(error.httpMethod).toEqual('get');
         expect(error.attempt).toEqual(1);
     });
+
+    test('should carry additional error data if provided', async () => {
+        const server = await mockServer.start();
```
barjin, line 68: Don't forget to `close()` what you open :) 

Btw in all the other tests, we're starting the server in `beforeAll()` (and closing it in `afterAll()`). I know that here it's only for this one test, but for the sake of consistency, can we move it? It's just a matter of time before we want to add a new test calling the mock server.

barjin (approved):
Aside from the test-related nit above, it seems pretty harmless to me 👍🏽

test/apify_api_error.test.js
gippy: Yea I was not sure what would be the correct approach here since I do think there will be other tests here in the future. But there is no harm in moving it there so I will do it, at least it will be consistent.

PR #576, opened 2024-08-09
feat: allow `unwind` param to `DatasetClient.listItems()` to be an array

fnesveda:
We're adding an option to unwind multiple fields when listing the dataset items. This adds an option to `DatasetClient` to pass those multiple fields in the `unwind` parameter.


drobnikj (changes requested):
Is it good approach to keep both string and sting[]? It can be not very clear for users as the rest params like omit, fields .. are available only string[].

I would prefer to keep unwind: string(to do not break) for some time but make it deprecate in typescript.

drobnikj (approved):

PR #578, opened 2024-08-16
feat(client): Add maxCostPerRun param

stetizu1:
Adds `maxCostPerRun` parameter to client - currently not doing anything, but not failing if it is received.

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -336,6 +336,8 @@ export interface ActorStartOptions {
      * Value can be accessed in actor run using `ACTOR_MAX_PAID_DATASET_ITEMS` environment variable.
      */
     maxItems?: number;
+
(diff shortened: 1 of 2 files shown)


mhamas (approved):
Please, have @B4nan or somebody from his team to review as well. Thanks

HonzaKirchner (approved):

barjin (changes requested):
I don't see anything wrong with the code, but I don't think we should merge this until it starts doing something :) Might be mighty confusing for users, despite the comment.

Any reason why you guys want this here this early?

barjin:
For bookkeeping: with @stetizu1 we agreed to modify only the `ow` validation and backend interaction while keeping the new parameter unlisted until it can be used by users.

barjin (approved):

PR #580, opened 2024-09-02
docs: use node 20 in the ci

B4nan:


.github/workflows/docs.yaml
```diff
@@ -22,10 +22,10 @@ jobs:
                 with:
                     token: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
 
-            -   name: Use Node.js 16
-                uses: actions/setup-node@v2-beta
```
janbuchar, line 26: Why a beta? :smile:

janbuchar (approved):

.github/workflows/docs.yaml
B4nan: no idea, this has been untouched for years apparently, it should be updated by renovate too

PR #595, opened 2024-10-20
feat: add user.updateLimits

### LINKED ISSUE #329: Add endpoints for user limits and usage
These are currently missing, so on Console, we have to make regular API calls without the client. It would be better to use the client.

MFori:
Resolves #329 - add `updateLimits` function to `UserClient`, which does `PUT: /v2/users/:userId/limits`
Api docs PR here: https://github.com/apify/openapi/pull/103
Endpoint is already implemented, only api docs and client was missing

### DIFF
--- a/src/resource_clients/user.ts
+++ b/src/resource_clients/user.ts
@@ -64,6 +64,19 @@ export class UserClient extends ResourceClient {
 
         return undefined;
     }
+
+    /**
+     * https://docs.apify.com/api/v2/#/reference/users/account-and-usage-limits
+     */
+    async updateLimits(limits: AccountLimitsUpdateOptions): Promise<void> {
+        const requestOpts: ApifyRequestConfig = {
+            url: this._url('limits'),
+            method: 'PUT',
+            params: this._params(),
+            data: limits,
+        };
+        await this.httpClient.call(requestOpts);
+    }
 }
 
 //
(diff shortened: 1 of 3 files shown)


src/resource_clients/user.ts
```diff
@@ -64,6 +64,19 @@ export class UserClient extends ResourceClient {
 
         return undefined;
     }
+
+    /**
+     * https://docs.apify.com/api/v2/#/reference/users/account-and-usage-limits
+     */
+    async updateLimits(limits: AccountLimitsUpdateOptions): Promise<void> {
```
tobice, line 71: (opt) 
```suggestion
    async updateLimits(options: AccountLimitsUpdateOptions): Promise<void> {
```

IMHO from language perspective, in `AccountLimitsUpdateOptions`, `Options` is the noun (_what it is_), and `Account` / `Limits` / `Update` serve as adjectives to make the noun more specific. So when you pass it around as a variable, I would call the variable primarily by what it is (i.e., _options_), and then add the adjectives to whatever extent is necessary. As the scope of this variable is tiny, just `options` is enough.

In other words, the parameter does not contain _limits_ (not all of them anyway), but _options to update limits_, so we should name the variable accordingly.

(This is a long comment for an optional nitpick :sweat_smile: But hopefully it makes sense where I'm coming from. It is also possible that this is a standard naming pattern that I don't know about, in which case feel free to discard the message.)

tobice (approved):
Just two naming suggestions, otherwise okay.

gippy (approved):

PR #596, opened 2024-10-23
ci: Automate the release process

janbuchar:
This is essentially the same thing that we do in apify/apify-sdk-python, apify/crawlee-python, etc., but for Javascript.

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -10,7 +10,7 @@ on:
 
 jobs:
     release_metadata:
-        if: "!startsWith(github.event.head_commit.message, 'docs') && !startsWith(github.event.head_commit.message, 'ci')"
+        if: "!startsWith(github.event.head_commit.message, 'docs') && !startsWith(github.event.head_commit.message, 'ci') && startsWith(github.repository, 'apify/')"
         name: Prepare release metadata
         runs-on: ubuntu-latest
         outputs:
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -17,16 +17,13 @@ jobs:
         runs-on: ubuntu-latest
         outputs:
             version_number: ${{ steps.release_metadata.outputs.version_number }}
-            tag_name: ${{ steps.release_metadata.outputs.tag_name }}
             changelog: ${{ steps.release_metadata.outputs.changelog }}
-            release_notes: ${{ steps.release_metadata.outputs.release_notes }}
         steps:
             -   uses: apify/workflows/git-cliff-release@main
                 name: Prepare release metadata
                 id: release_metadata
                 with:
-                    release_type: ${{ inputs.release_type }}
-                    custom_version: ${{ inputs.custom_version }}
+                    release_type: prerelease
 
     wait_for_checks:
         name: Wait for code checks to pass
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -5,9 +5,8 @@ on:
     push:
         branches:
             - master
-            # This will allow releasing beta versions from non-master releases.
-            # TODO: We need to release this version with other tags than beta. It can mess up the beta version as it will override it with older code.
-            - version/**
+        tags-ignore:
+            - "**" # Ignore all tags to prevent duplicate builds when tags are pushed.
 
 jobs:
     release_metadata:
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -55,7 +55,7 @@ jobs:
                     node-version: 22
 
             -   name: Update package version in package.json
-                run: npm version --no-git-tag-version ${{ needs.release_metadata.outputs.version_number }}
+                run: npm version --no-git-tag-version --allow-same-version ${{ needs.release_metadata.outputs.version_number }}
 
             -   name: Update CHANGELOG.md
                 uses: DamianReeves/write-file-action@master
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -55,7 +55,7 @@ jobs:
                     node-version: 22
 
             -   name: Update package version in package.json
-                run: npm version ${{ needs.release_metadata.outputs.version_number }}
+                run: npm version --no-git-tag-version ${{ needs.release_metadata.outputs.version_number }}
 
             -   name: Update CHANGELOG.md
                 uses: DamianReeves/write-file-action@master
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -1,28 +1,17 @@
-name: Create a release
+name: Create a pre-release
 
 on:
-    # Trigger a stable version release via GitHub's UI, with the ability to specify the type of release.
-    workflow_dispatch:
-        inputs:
-            release_type:
-                description: Release type
-                required: true
-                type: choice
-                default: auto
-                options:
-                    - auto
-                    - custom
-                    - patch
-                    - minor
-                    - major
-            custom_version:
-                description: The custom version to bump to (only for "custom" type)
-                required: false
-                type: string
-                default: ""
+    # Push to master will deploy a beta version
+    push:
+        branches:
+            - master
+            # This will allow releasing beta versions from non-master releases.
+            # TODO: We need to release this version with other tags than beta. It can mess up the beta version as it will override it with older code.
+            - version/**
 
 jobs:
     release_metadata:
+        if: "!startsWith(github.event.head_commit.message, 'docs') && !startsWith(github.event.head_commit.message, 'ci')"
         name: Prepare release metadata
         runs-on: ubuntu-latest
         outputs:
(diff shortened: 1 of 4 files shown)


janbuchar:
@vladfrangu I'm tagging you because I'll try to do the same thing to CLI next. What do you think?

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
+name: Create a pre-release
+
+on:
+    # Push to master will deploy a beta version
+    push:
+        branches:
+            - master
+            # This will allow releasing beta versions from non-master releases.
+            # TODO: We need to release this version with other tags than beta. It can mess up the beta version as it will override it with older code.
```
vladfrangu, line 9: What does this mean exactly?

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
...
+            - version/**
+
+jobs:
+    release_metadata:
+        if: "!startsWith(github.event.head_commit.message, 'docs') && !startsWith(github.event.head_commit.message, 'ci')"
```
vladfrangu, line 14: Lets add in a check that it runs on our org (so forked users dont have this try to needlessly run)

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
...
+                    message: "chore(release): Update changelog and package version [skip ci]"
+
+    publish_to_npm:
+        name: Publish to NPM
+        needs: [ release_metadata, wait_for_checks ]
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: actions/checkout@v4
+                with:
+                    ref: ${{ needs.update_changelog.changelog_commitish }}
```
vladfrangu, line 81: Sooo we always push a changelog update every beta release... gottt it

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
...
+            -   name: Use Node.js 22
+                uses: actions/setup-node@v4
+                with:
+                    node-version: 22
+                    cache: 'npm'
+                    cache-dependency-path: 'package-lock.json'
```
vladfrangu, line 87: pretty sure this can be removed as its the default

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
...
+            -   name: Install dependencies
+                run: |
+                    echo "access=public" >> .npmrc
+                    echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" >> .npmrc
+                    npm ci
```
vladfrangu, line 92: Can we make a reusable step for setting up npmrc / yarnrc / etc with auth tokens? 🙏 (not in this PR, but in general/future)

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,105 @@
+name: Create a pre-release
```
vladfrangu, line 1: This flow should really have a concurrency group and either
- let only the latest commit run
- wait for existing runs before proceeding

.github/workflows/release.yaml
vladfrangu, line 1: Are we sure this will check the `check` flow for results when calling via workflow dispatch? Or is it smart enough to correlate with the git sha of the commit we have on the branch when we run it?

vladfrangu (dismissed):
Overall tho I am _loving_ what I see here 🎉, this has been needed for faaar too long

.github/workflows/pre_release.yaml
janbuchar: no idea honestly, that was already in the current version :smile:

.github/workflows/release.yaml
janbuchar: I'm pretty sure. It checks `${{ github.ref }}`, which should contain e.g., `master` when you trigger a release with `workflow_dispatch`.

.github/workflows/pre_release.yaml
janbuchar: Yeah, excellent point. I've seen many different takes on this in our projects. Do you know which is the favored approach?

.github/workflows/pre_release.yaml
janbuchar: Just the path, or `cache: npm` as well?

.github/workflows/pre_release.yaml
janbuchar: Not sure I understand, but I admit I didn't consider forks too hard. Can you elaborate please?

.github/workflows/pre_release.yaml
janbuchar: Well, I think that the whole "publish to npm" thing could be a separate workflow/action that could also contain that `before-beta-release.js` script.

.github/workflows/pre_release.yaml
B4nan: i would keep it, since we will be porting this to places that use yarn too, and it will be easier to rewrite just the value instead of googling the key

.github/workflows/pre_release.yaml
B4nan: I think it makes sense to finish the publishing for canary builds that already started. We do cancel the tests workflow, and I'd say that's fine, but if we already started with the canary build, we should finish that. So I would go with the second option personally.

janbuchar: Sure, makes sense. Do we already do that in some of our workflows?

B4nan: Probably not, I honestly don't know how this can be done, I only ever dealt with canceling, not the other way around. @vladfrangu do you have something specific on your mind outside of removing the cancel action (which I would rather not)?

vladfrangu: I'll get that to you in a few!

vladfrangu: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency

Just set cancel-in-progress to false and pick a group (probs unique)

B4nan: Nice, sounds like we could also replace the `styfle/cancel-workflow-action` action we use everywhere with this native solution then?

.github/workflows/pre_release.yaml
vladfrangu: Basically adding `github.repository_owner == 'apify'` in the if-check

.github/workflows/pre_release.yaml
vladfrangu: Its likely, yes! But we'd need to be careful, for some flows we should still use the concurrency with cancel-in-progress set to true (like the testing on platform ones)

.github/workflows/pre_release.yaml
janbuchar: Right, because when somebody makes a fork, it might contain a master branch and you don't want to release from that. Got it.

B4nan: Hmm, do we deal with this anywhere else? I don't think so, and I've never heard of some failed release pipeline from a fork, are we sure this is valid problem? Either way, the fork won't have secrets, so we don't have to be afraid of that much, but I wonder if it can actually happen.

.github/workflows/pre_release.yaml
B4nan: If it makes no sense to nobody (I also have no idea, feel free to check git blame), let's just drop that comment right ahead.

.github/workflows/pre_release.yaml
vladfrangu: Its mostly so we don't needlessly run steps that will fail anyways on forks (all 5 of them or however many we have :D)

B4nan: I don't follow, what 5 forks we have?

My question still holds, where are we dealing with this? Because I am not aware of a single workflow where we would check the org name.

I am not saying its a bad idea, but it feels weird that it only pops now, if it would be an actual problem, how come we haven't seen that years ago?

.github/workflows/pre_release.yaml
B4nan: yeah, lets leave that for later, we have better things to do than breaking things that work :D

.github/workflows/pre_release.yaml
janbuchar: I'm not sure that we did pre-releases on push to master before - that might be the reason we're only dealing with this now.

.github/workflows/pre_release.yaml
janbuchar: I'd also remove the line that follows it :slightly_smiling_face:

.github/workflows/pre_release.yaml
B4nan: True, but the previous version was adding git tags for each canary build, which isn't that different.

Anyway, let's do it, although I would love to see a better solution than hardcoding org name (but lets not waste much time on this).

janbuchar: You don't need NPM credentials to make a git tag locally though :slightly_smiling_face: Also I'm not sure which name the default branch for a new fork receives, but I'd wager it won't be `master`.

All aside, just checking if this happens under `apify/` is just fine IMO.

.github/workflows/pre_release.yaml
janbuchar: Anyway, https://github.com/apify/workflows/issues/145

B4nan (approved):

PR #602, opened 2024-10-31
feat: allow appending custom parts to the user agent

B4nan:
Related https://github.com/apify/apify-sdk-js/issues/331

### DIFF
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -53,7 +53,7 @@ export class ApifyClient {
             requestInterceptors: ow.optional.array,
             timeoutSecs: ow.optional.number,
             token: ow.optional.string,
-            userAgentSuffix: ow.any(ow.optional.string, ow.optional.array.ofType(ow.string)),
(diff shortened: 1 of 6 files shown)


src/apify_client.ts
```diff
@@ -53,6 +53,7 @@ export class ApifyClient {
             requestInterceptors: ow.optional.array,
             timeoutSecs: ow.optional.number,
             token: ow.optional.string,
+            userAgentSuffix: ow.any(ow.optional.string, ow.optional.array.ofType(ow.string)),
```
vladfrangu, line 56: Assertion doesn't match type

vladfrangu (changes requested):

src/apify_client.ts
B4nan: I swear I started with modifying the type and only then modified this, how come the changes were lost :D

vladfrangu (approved):

PR #604, opened 2024-11-01
fix: allow `null` when updating dataset/kvs/rq `name`

B4nan:
Related https://github.com/apify/apify-cli/pull/685


janbuchar (approved):
LGTM

PR #612, opened 2024-11-28
fix(actor): correctly set type for ActorTaggedBuilds

metalwarrior665:
I have no idea why we had such a wrong type there.

1. There doesn't have to be `latest` tag
2. Tags are arbitrary strings

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -268,9 +268,7 @@ export interface ActorExampleRunInput {
     contentType: string;
 }
 
-export interface ActorTaggedBuilds {
-    latest: ActorTaggedBuild;
-}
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

src/resource_clients/actor.ts
```diff
@@ -268,9 +268,7 @@ export interface ActorExampleRunInput {
     contentType: string;
 }
 
-export interface ActorTaggedBuilds {
-    latest: ActorTaggedBuild;
-}
+export type ActorTaggedBuilds = Record<string, ActorTaggedBuild>
```
B4nan, line 271: not sure why this is not checked via eslint, there should be a semicolon at the end 🙃

PR #613, opened 2024-12-03
feat: Add `charge` method to the run client for "pay per event"

Jkuzz:
Resolves https://github.com/apify/apify-core/issues/18592 by adding the PPE charge endpoint to JS client.

The idempotency key creation was taken from an actor by the store team.
Issue to add docs URL https://github.com/apify/apify-client-js/issues/614

### DIFF
--- a/src/resource_clients/run.ts
+++ b/src/resource_clients/run.ts
@@ -154,7 +154,7 @@ export class RunClient extends ResourceClient {
         const count = options.count ?? 1;
         /** To avoid duplicates during the same milisecond, doesn't need to by crypto-secure. */
         const randomSuffix = (Math.random() + 1).toString(36).slice(3, 8);
-        const idempotencyKey = options.idempotencyKey ?? `${this.id}-${options.eventName}-${randomSuffix}`;
+        const idempotencyKey = options.idempotencyKey ?? `${this.id}-${options.eventName}-${Date.now()}-${randomSuffix}`;
 
         const request: AxiosRequestConfig = {
             url: this._url('charge'),
--- a/src/resource_clients/run.ts
+++ b/src/resource_clients/run.ts
@@ -261,6 +261,7 @@ export interface RunResurrectOptions {
 }
 
 export type RunChargeOptions = {
+    /** Name of the event to charge. Must be defined in the Actor's pricing info else the API will throw. */
     eventName: string;
     /** Defaults to 1 */
     count?: number;
--- a/src/resource_clients/run.ts
+++ b/src/resource_clients/run.ts
@@ -13,6 +13,7 @@ import {
     parseDateFields,
     cast,
 } from '../utils';
+import { ApifyResponse } from '../http_client';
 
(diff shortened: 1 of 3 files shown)


src/resource_clients/run.ts
```diff
@@ -138,6 +141,35 @@ export class RunClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    /**
+     * TODO: docs url
+     * https://github.com/apify/apify-client-js/issues/614
```
B4nan, line 146: can we add the URL here right ahead? even if it doesn't exist yet, feels better than publishing with a TODO

B4nan (approved):
you can ignore the failing docs build, its not detecting the links inside parens correctly

src/resource_clients/run.ts
Jkuzz: I don't have the URL. I'll ask Mat'o what's the status on the endpoint docs 👍

B4nan: I am rather sure we can guess it before it exists, its just another endpoint, the URLs for them are not random.

src/resource_clients/run.ts
```diff
@@ -138,6 +141,34 @@ export class RunClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    /**
+     * https://docs.apify.com/api/v2#/reference/actor-runs/charge-run/charge-run
+     */
+    async charge(options: RunChargeOptions): Promise<ApifyResponse<Record<string, never>>> {
```
barjin, line 147: Out of curiosity - what's the expected return type of this method? I see that most of the `RunClient` methods return `ActorRun` object (which reflects the changes made by the call). 

Is this return type (`ApifyResponse<Record<string, never>>`) a temporary thing before we figure out the actual API response shape, or is this the final thing?

Jkuzz: We discussed this in person. The endpoint uses a memorized version of the run from the database for performance reasons, as the run's pricing model cannot change during the run. Therefore we cannot return the modified run object and the only important output is the response status code.

src/resource_clients/run.ts
```diff
@@ -138,6 +141,34 @@ export class RunClient extends ResourceClient {
...
+        ow(options, ow.object.exactShape({
+            eventName: ow.string,
+            count: ow.optional.number,
+            idempotencyKey: ow.optional.string,
+        }));
+
+        const count = options.count ?? 1;
+        const idempotencyKey = options.idempotencyKey ?? `${this.id}-${options.eventName}-${Date.now()}`;
```
fnesveda, line 155: Same note here as in the Python client, the idempotency key has to be a unique string. This way, if someone charged for two occurences of the same event in the same millisecond, only one of the charges would go through.

https://github.com/apify/apify-client-python/pull/304#pullrequestreview-2480812542

fnesveda (changes requested):
One problem copied over from the Python client review

fnesveda (approved):

PR #621, opened 2025-01-01
feat: add dataset.statistics

MFori:
Part of https://github.com/apify/apify-core/issues/18807 - Implementation of new API endpoint `v2/datasets/{datasetId}/statistics`

### DIFF
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -4,7 +4,7 @@ import { ApifyApiError } from '../apify_api_error';
 import { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceClient } from '../base/resource_client';
 import { ApifyRequestConfig, ApifyResponse } from '../http_client';
-import {cast, catchNotFoundOrThrow, PaginatedList, pluckData} from '../utils';
+import { cast, catchNotFoundOrThrow, PaginatedList, pluckData } from '../utils';
 
 export class DatasetClient<
     Data extends Record<string | number, any> = Record<string | number, unknown>,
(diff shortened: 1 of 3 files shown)


src/resource_clients/dataset.ts
```diff
@@ -130,6 +131,24 @@ export class DatasetClient<
...
+        const requestOpts: ApifyRequestConfig = {
+            url: this._url('statistics'),
+            method: 'GET',
+            params: this._params(),
+        };
+        try {
+            const response = await this.httpClient.call(requestOpts);
+            return cast(pluckData(response.data));
+        } catch (err) {
+            catchNotFoundOrThrow(err as ApifyApiError);
```
drobnikj, line 147: I'm not sure about this, should be the same as listItems and return not found error once the dataset does not exist?

drobnikj (commented):
Looks fine, I'm not sure about catchNotFoundOrThrow, I would expect that the method throw, why did you use it?

src/resource_clients/dataset.ts
MFori: I agree it makes more sense to me too, but I got inspired by `DatasetClient.get()` which doesn't throw if dataset not exists 🤔 (which I would expect to also throw)

B4nan: I would also prefer throwing when things are not there, but we should be consistent, and it looks like most methods use the `catchNotFoundOrThrow` approach.

drobnikj: `catchNotFoundOrThrow` feels confusing, if you get undefinied, you have no idea if the dataset does not exist or statistics missing.

drobnikj (approved):
I would prefere throw over catchNotFoundOrThrow, see discussion. But pre aproving as even catchNotFoundOrThrow makes sence if we want to be consistent.

PR #623, opened 2025-01-06
fix: Add ActorRunPricingInfo type

janbuchar:


janbuchar:
We should also add `chargedEventCounts`.

src/resource_clients/actor.ts
```diff
@@ -459,3 +462,50 @@ export interface ActorDefinition {
...
+}
+
+export type FreeActorPricingInfo = CommonActorPricingInfo & {
+    pricingModel: 'FREE';
+}
+
+export type FlatPricePerMonthActorPricingInfo = CommonActorPricingInfo & {
+    pricingModel: 'FLAT_PRICE_PER_MONTH';
+    /** for how many days this Actor can be used for free in trial period */
+    trialMinutes?: number;
```
barjin, line 485: Days (jsdoc) or minutes (variable name)?

src/resource_clients/actor.ts
```diff
@@ -459,3 +462,50 @@ export interface ActorDefinition {
...
+    eventTitle: string;
+    eventDescription?: string;
+}>
+
+export type PricePerEventActorPricingInfo = CommonActorPricingInfo & {
+    pricingModel: 'PAY_PER_EVENT';
+    pricingPerEvent: {
+        actorChargeEvents: ActorChargeEvents
+    };
+    minimalMaxTotalChargeUsd?: number;
```
barjin, line 508: I know the reasoning behind this, since I overheard some Slack conversation, but other ppl might not.

Can we add jsdocs for this (and other properties as well)? 

Also, I know that we won't solve this here, but how about we call it `minimalChargeLimitUsd` (or even `minTotalUsd`)? *Minimal maximum total charge in US dollars* sounds like something straight outta Dr. Seuss book.

barjin (commented):
Looks alright to me, I'd maybe capitalize all the JSDocs (and add more details here and there).

src/resource_clients/actor.ts
```diff
@@ -459,3 +462,50 @@ export interface ActorDefinition {
...
+type CommonActorPricingInfo = {
+    /** In [0, 1], fraction of pricePerUnitUsd that goes to Apify */
+    apifyMarginPercentage: number;
+    /** When this pricing info record has been created */
+    createdAt: Date;
+    /** Since when is this pricing info record effective for a given Actor */
+    startedAt: Date;
+    notifiedAboutFutureChangeAt?: Date;
+    notifiedAboutChangeAt?: Date;
+    reasonForChange?: string;
```
barjin, line 475: Can we get docs for these properties as well? I think could guess the meaning of `startedAt` or `apifyMarginPercentage`, but I'm honestly getting lost with `notifiedAboutFutureChangeAt` or `reasonForChange`.

Again, I know this is a conversation for a different thread, but `apifyMarginPercentage` / `createdAt` / `startedAt` make sense as "Pricing" properties... the other three, IMO not so much.

src/resource_clients/actor.ts
```diff
@@ -459,3 +462,50 @@ export interface ActorDefinition {
...
+}
+
+export type FreeActorPricingInfo = CommonActorPricingInfo & {
+    pricingModel: 'FREE';
```
barjin, line 479: Maybe a string enum for the `pricingModel` variants? I've got no hard opinion here though

barjin (commented):
few more nits

src/resource_clients/actor.ts
janbuchar: This is copied more or less as is from the implementation. It doesn't feel right to start writing our own version of the docs here... In a better world, these types would be generated from the OpenAPI spec, but that's a pipe dream, I guess.

barjin (approved):
As long as you are sure about the actual types and response shapes, lgtm 🚀

B4nan (changes requested):
i would prefer interfaces over types, simply because of how they are rendered in the API docs

interface has its own page, all properties listed, inheritance is resolved: https://docs.apify.com/api/client/js/reference/interface/Actor
type is just listed on one huge page and you only see the implementation basically: https://docs.apify.com/api/client/js/reference#ActorCollectionListResult

src/resource_clients/actor.ts
janbuchar: I forwarded the issue :shrug:

janbuchar:
> i would prefer interfaces over types, simply because of how they are rendered in the API docs

sure thing, you got it

B4nan (approved):

PR #624, opened 2025-01-09
fix: change type `Build.actorDefinitions` to `Build.actorDefinition`

jirispilka:
It should be `actorDefinition` instead of `actorDefinitions`

I've checked in db and by calling:
```
curl -L 'https://api.apify.com/v2/actor-builds/Nxeu6qxhoEs90eHIH' \
-H 'Accept: application/json'
```
Response
```
    "gitBranchName": "master",
    "actorDefinition": {
      "actorSpecification": 1,
      "name": "rag-web-browser",
      "title": "RAG Web browser",
```

### DIFF
--- a/src/resource_clients/build.ts
+++ b/src/resource_clients/build.ts
@@ -123,7 +123,7 @@ export interface Build {
     usage?: BuildUsage;
     usageTotalUsd?: number;
     usageUsd?: BuildUsage;
-    actorDefinitions?: ActorDefinition;
+    actorDefinition?: ActorDefinition;
 }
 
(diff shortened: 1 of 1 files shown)


janbuchar:
The docs agree with this - https://docs.apify.com/api/v2-new/actor-build-get and the Python client apparently doesn't even support this endpoint, so there's no need to make this change elsewhere.

janbuchar (approved):

drobnikj (approved):
Thanks for the fix, probably just my typo.

PR #630, opened 2025-01-27
fix(test): fix flaky test

drobnikj:
This reverts commit 6175cfb.

### DIFF
--- a/test/request_queues.test.js
+++ b/test/request_queues.test.js
@@ -331,6 +331,21 @@ describe('Request Queue methods', () => {
             );
         });
 
+        test('batchAddRequests() propagates forefront', async () => {
+            const queueId = 'some-id';
+            const options = { forefront: true };
+            const requests = new Array(10).fill(0).map((_, i) => ({ url: `http://example.com/${i}` }));
+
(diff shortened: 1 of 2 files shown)


drobnikj:
Looks it generates other problem 🤯  let me check it.

test/request_queues.test.js
```diff
@@ -492,5 +463,35 @@ describe('Request Queue methods', () => {
             expect(browserRes.items).toEqual(browserRequests);
             validateRequest({ limit: maxPageLimit }, { queueId });
         });
+
+        // NOTE: This test needs to be last otherwise it will break other tests
```
B4nan, line 467: it would be much better to remove the "dependency", tests should be ideally isolated. isn't it enough to use a different queue ID?

drobnikj: The problem is how the mock server is written
Checking request from mock server is done very basic using pop() method from list, see
https://github.com/apify/apify-client-js/blob/master/test/mock_server/server.js#L48
so changing ID does not help.

B4nan:
shall we merge this or why did you keep it as a draft?

drobnikj:
let's merge this, seems like a good enough fix.

B4nan (approved):

PR #633, opened 2025-02-03
docs: add v2.10 and v2.11 version snapshots

B4nan:


danpoletaev (approved):

website/versioned_docs/version-2.11/api-packages.json
```diff
@@ -1 +1 @@
-[{"entryPoints":{"index":{"label":"Index","path":"src/index.ts"}},"packageRoot":"/Users/adamek/htdocs/apify/apify-client-js","packagePath":".","packageSlug":".","packageName":"apify-client","packageVersion":"2.11.2"}]
\ No newline at end of file
+[{"entryPoints":{"index":{"label":"Index","path":"src/index.ts"}},"packageRoot":"/home/jindrichbar/Desktop/apify/apify-client-js","packagePath":".","packageSlug":".","packageName":"apify-client","packageVersion":"2.11.0"}]
```
B4nan, line 1: huh, how come this is an absolute path. the old snapshots have all relative path in here

barjin: Damn, could this be the issue? I've been trying to make it work the whole afternoon, no dice... then I regenerated the snapshots and it started working.

The true definition of _it worked on my computer_.

barjin:
It seems that the newer versions (>4) of `@apify/docusaurus-plugin-typedoc-api` don't generate valid API snapshots. I have no idea what the difference is - I just generated the snapshots for the `2.10` and `2.11` with older version to unblock this PR.

I'll deal with the `docusaurus-plugin-typedoc-api` issue separately there.

website/versioned_docs/version-2.11/api-packages.json
barjin: (btw removing the `packageRoot` field manually or replacing it with a relative path did not help).

B4nan:
Thanks!

PR #668, opened 2025-04-09
feat: add defaultBuild method

danpoletaev:
This PR adds `defaultBuild()` method to Actor client

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -163,11 +163,11 @@ export class ActorClient extends ResourceClient {
     /**
      * https://docs.apify.com/api/v2/act-build-default-get
      */
-    async defaultBuild(): Promise<Build> {
+    async defaultBuild(options: Pick<ActorStartOptions, 'waitForFinish'> = {}): Promise<Build> {
         const response = await this.httpClient.call({
             url: this._url('builds/default'),
             method: 'GET',
-            params: this._params(),
+            params: this._params(options),
         });
 
         return cast(parseDateFields(pluckData(response.data)));
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -161,6 +161,19 @@ export class ActorClient extends ResourceClient {
     }
 
     /**
+     * https://docs.apify.com/api/v2/act-build-default-get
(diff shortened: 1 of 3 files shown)


src/resource_clients/actor.ts
```diff
@@ -160,6 +160,19 @@ export class ActorClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    /**
+     * https://docs.apify.com/api/v2/act-build-default-get
+     */
+    async defaultBuild(): Promise<Build> {
```
barjin, line 166: I see the API docs discuss the `waitForFinish` parameter. Is there any chance we can support it here, too?

image

barjin (changes requested):
Aside from the missing parameter, it seems quite straightforward. Thanks!

src/resource_clients/actor.ts
danpoletaev: Good point 👍 
I've added it to params

src/resource_clients/actor.ts
```diff
@@ -160,6 +160,19 @@ export class ActorClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    /**
+     * https://docs.apify.com/api/v2/act-build-default-get
+     */
+    async defaultBuild(options: Pick<ActorStartOptions, 'waitForFinish'> = {}): Promise<Build> {
```
barjin, line 166: We could use the `BuildClientGetOptions` interface [from here](https://github.com/apify/apify-client-js/blob/6e520aef6f19f0905ea3dabcfaab7dbf12ab28fa/src/resource_clients/build.ts#L99C18-L99C39), right? Do I understand correctly this endpoint is basically like `GET v2/actor-builds/:buildId`, just without the explicit build ID (with the Actor id instead)?

danpoletaev: Oh yeah, that's much better!

Yeah, it behaves same as `/v2/actor-builds/:buildId`, but it is in acts namespace:
`/v2/acts/:actorId/builds/:buildId` or `/v2/acts/:actorId/builds/default` for default build of the actor

PR #669, opened 2025-04-11
feat: add support for general resource access

tobice:
We plan to allow users to configure through API & clients how their storages and runs can be accessed — whether just the ID / name is sufficient to access the storage. This is controlled through the `generalAccess` resource property, which overrides account setting for the resource.

Full context in https://github.com/apify/apify-core/issues/19012.

This PR updates the client so that it can be used to get and update general access on datasets, key-value stores, request queues, and runs.

```typescript
const run = client.run(runId);

// Set general access to ANYONE_WITH_ID_CAN_READ which overrides the account setting.
await run.update({ generalAccess: RUN_GENERAL_ACCESS.ANYONE_WITH_ID_CAN_READ });

// Reset the general access so that it follows the account setting.
await run.update({ generalAccess: null });
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
 		"build:browser": "webpack"
 	},
 	"dependencies": {
-		"@apify/consts": "^2.39.0",
+		"@apify/consts": "^2.40.0",
 		"@apify/log": "^2.2.6",
 		"@crawlee/types": "^3.3.0",
 		"agentkeepalive": "^4.2.1",
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
 		"build:browser": "webpack"
 	},
 	"dependencies": {
-        "@apify/consts": "^2.39.0",
+		"@apify/consts": "^2.39.0",
 		"@apify/log": "^2.2.6",
 		"@crawlee/types": "^3.3.0",
 		"agentkeepalive": "^4.2.1",
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
 		"build:browser": "webpack"
 	},
 	"dependencies": {
-		"@apify/consts": "^2.25.0",
+        "@apify/consts": "^2.39.0",
 		"@apify/log": "^2.2.6",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
 		"build:browser": "webpack"
 	},
 	"dependencies": {
-		"@apify/consts": "^2.25.0",
+        "@apify/consts": "^2.39.0",
```
B4nan, line 62: not sure why we have tabs in here, but we should stay consistent

src/resource_clients/key_value_store.ts
```diff
@@ -18,6 +18,7 @@ import {
     parseDateFields,
     pluckData,
 } from '../utils';
+import { STORAGE_GENERAL_ACCESS } from '@apify/consts';
```
B4nan, line 21: this should be higher, before the relative imports. 

i'll take a look at the eslint config, it should have caught this, i think i know why its not working

src/resource_clients/run.ts
```diff
@@ -14,6 +14,7 @@ import {
     parseDateFields,
     cast,
 } from '../utils';
+import { RUN_GENERAL_ACCESS } from '@apify/consts';
```
B4nan, line 17: same here

B4nan (changes requested):
just a few code style nits

package.json
tobice: Huh, tabs? :smile: Fixed :shrug:

B4nan:
#670 will fix the eslint config

src/resource_clients/key_value_store.ts
```diff
@@ -4,6 +4,7 @@ import ow from 'ow';
 import type { JsonValue } from 'type-fest';
 
 import log from '@apify/log';
+import { STORAGE_GENERAL_ACCESS } from '@apify/consts';
```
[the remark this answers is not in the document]
tobice: Fixed eslint :+1:

PR #671, opened 2025-04-11
chore: update eslint, adopt prettier and rsbuild

### LINKED ISSUE #607: Update eslint to v9, use new flat Apify eslint config
We should update eslint to v9 along with the new flat Apify eslint config that supports it.

See [notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be](https://www.notion.so/apify/New-linting-and-formatting-configuration-across-Apify-0b62890892cd4b71b4fe2dee9291b4af?pvs=4#6152e086487441f6b974e6fbda9614be) for a detailed guide.

B4nan:
- update of eslint config
- replacement of webpack with rsbuild for the browser bundle (with support for `node:` imports that we now have in the shared repo)
- adopts prettier and reformats everything, including the docs and workflows
- coverage is broken because of the changes in browser bundle, didnt find a way to make it fly, but i guess its fine for the client and we can just disable those?

Closes #607


src/apify_client.ts
```diff
@@ -1,7 +1,7 @@
 import type { SetStatusMessageOptions } from '@crawlee/types';
 import ow from 'ow';
 
-import { ME_USER_NAME_PLACEHOLDER, ACTOR_ENV_VARS } from '@apify/consts';
+import { ACTOR_ENV_VARS,ME_USER_NAME_PLACEHOLDER } from '@apify/consts';
```
B4nan, line 4: Duh, the fixer is messing up white spaces, seen that before.

I guess I'll add prettier right ahead instead of dealing with this by hand.

B4nan (commented):
this will be more fun, moving to draft

src/apify_client.ts
tobice: Alright, I can see this is in draft, so withholding my review.

jirimoravcik: Hmm, this is dumb, haven't seen it cause problems in any other projects.

B4nan: It happened on crawlee, sdk as well.

B4nan:
@jirimoravcik @tobice this is way too huge to be actually reviewed because of the prettier reformatting everything, including the docs, but i still need a checkmark 🙃 

The only thing that is not working the same as before is the code coverage, since we test on the compiled client-side bundle, but i hope that shouldn't matter much, i don't think anybody was actually paying attention to it.

jirimoravcik (approved):

PR #677, opened 2025-04-22
fix: defaultBuild() returns BuildClient

danpoletaev:
This PR updates the Actor.defaultBuild() method to return a BuildClient instance instead of a raw response object.


barjin (approved):
imo looking good, thank you!

PR #678, opened 2025-04-24
ci: Fix running code checks in release workflow

janbuchar:


barjin (approved):
Looking solid, thank you!

I guess `wait-on-check-action` never invoked `check.yaml` right? Curious we haven't noticed it earlier :sweat_smile:

B4nan (approved):

janbuchar:
> Looking solid, thank you!
> 
> I guess `wait-on-check-action` never invoked `check.yaml` right? Curious we haven't noticed it earlier 😅

Well, if the checks passed for the last commit on the target branch, everything worked fine... And those are invoked automatically after merging a PR. Trouble starts when the head commit is a `[ci skip]` one - the wait-for-checks action cannot handle that. And because of the way we update changelogs, there are many such commits :see_no_evil:

PR #683, opened 2025-04-27
fix(actor): add missing 'pricingInfos' field to Actor object

metalwarrior665:
The object should be identical as the one in the run:

https://docs.apify.com/api/v2/act-get
vs
https://docs.apify.com/api/v2/act-run-get


barjin (approved):
Looks solid to me, thanks!

PR #691, opened 2025-05-09
fix: add missing 'effectivePlatformFeatures', 'createdAt', 'isPaying' to User interface

metalwarrior665:
Prerequisite https://github.com/apify/apify-docs/pull/1569 was merged

Questions:
1. The string -> Date conversion happens automatically for all types, right? So no need to do anything?
2. Do we use explicit `T | null` anywhere if the API returns the type or null? Seems we like to do `T?` rather

### DIFF
--- a/src/resource_clients/user.ts
+++ b/src/resource_clients/user.ts
@@ -103,7 +103,7 @@ export interface User {
     email?: string;
     proxy?: UserProxy;
     plan?: UserPlan;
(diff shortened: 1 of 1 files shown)


B4nan (approved):
looks good, but please ensure the nulls are correct (so if the API can return null somewhere, we have it on type level, if not and the API would instead omit the property, it shouldn't be there)

metalwarrior665:
I tested the Date conversion locally and double-checked the explicit nulls

PR #692, opened 2025-05-13
feat: allow overriding timeout of `KVS.setRecord` calls

B4nan:
Related: https://apify.slack.com/archives/C0L33UM7Z/p1747043120071539


src/resource_clients/key_value_store.ts
```diff
@@ -177,7 +177,8 @@ export class KeyValueStoreClient extends ResourceClient {
         );
 
         const { key } = record;
-        let { value, contentType } = record;
+        // eslint-disable-next-line prefer-const
+        let { value, contentType, timeoutSecs, doNotRetryTimeouts } = record;
```
tobice, line 181: (nit) This is not really part of the record. Wouldn't it be cleaner to pass this inside an optional `options` param?

tobice (changes requested):

src/resource_clients/key_value_store.ts
B4nan: I knew you won't like this bit 🤣

B4nan: I'll refactor it here, but in crawlee I'd rather keep a single options object, feels weird to have two options objects next to each other.

https://github.com/apify/crawlee/blob/master/packages/core/src/storages/key_value_store.ts#L328

tobice (approved):

PR #695, opened 2025-05-13
Move prettier into devDependencies

hudson-worden:
prettier is just for formatting - so it should go into devDependencies. This takes out the transitive dependency on prettier `^3.5.3` which occurs b/c it is not in devDependencies.


vladfrangu (approved):

B4nan (approved):
thanks!

PR #704, opened 2025-06-09
fix: Use appropriate timeouts

### LINKED ISSUE #685: Refactor client timeouts - parity with Python implemenation
Due to some API calls having too long default timeout, the crawlers can "freeze" for several minutes waiting for api call. In Python version this was fixed by reducing the default timeout for some endpoints and increasing the timeout in case of retries.

See Python implementation for context: https://github.com/apify/apify-client-python/pull/384
Original issue: https://github.com/apify/crawlee-python/issues/1132

janbuchar:
- closes #685
- modelled after https://github.com/apify/apify-client-python/pull/384

### DIFF
--- a/src/base/resource_client.ts
+++ b/src/base/resource_client.ts
@@ -13,21 +13,21 @@ import { ApiClient } from './api_client';
  */
 const MAX_WAIT_FOR_FINISH = 999999;
 
-export const SMALL_TIMEOUT_SECS = 5; // For fast and common actions. Suitable for idempotent actions.
-export const MEDIUM_TIMEOUT_SECS = 30; // For actions that may take longer.
-export const DEFAULT_TIMEOUT_SECS = 360; // 6 minutes
+export const SMALL_TIMEOUT_MILLIS = 5 * 1000; // For fast and common actions. Suitable for idempotent actions.
+export const MEDIUM_TIMEOUT_MILLIS = 30 * 1000; // For actions that may take longer.
+export const DEFAULT_TIMEOUT_MILLIS = 360 * 1000; // 6 minutes
 
 /**
  * Resource client.
  * @private
  */
 export class ResourceClient extends ApiClient {
-    protected async _get<T, R>(options: T = {} as T, timeoutSecs?: number): Promise<R | undefined> {
+    protected async _get<T, R>(options: T = {} as T, timeoutMillis?: number): Promise<R | undefined> {
         const requestOpts: ApifyRequestConfig = {
             url: this._url(),
             method: 'GET',
             params: this._params(options),
-            timeout: timeoutSecs !== undefined ? timeoutSecs * 1000 : undefined,
+            timeout: timeoutMillis,
         };
         try {
             const response = await this.httpClient.call(requestOpts);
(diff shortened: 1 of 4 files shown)


src/http_client.ts
```diff
@@ -152,13 +152,21 @@ export class HttpClient {
...
                     // Handling redirects is not possible without buffering - part of the stream has already been sent and can't be recovered
                     // when server sends the redirect. Therefore we need to override this in Axios config to prevent it from buffering the body.
                     // see also axios/axios#1045
                     config = { ...config, maxRedirects: 0 };
                 }
+
+                // Increase timeout with each attempt. Max timeout is bounded by the client timeout.
+                config.timeout = Math.min(
+                    this.timeoutMillis,
+                    config.timeout ?? this.timeoutMillis * 2 ** (attempt - 1),
```
barjin, line 167: Nullish coalescing has lower precedence than multiplication. 

This would evaluate to `config.timeout ?? (this.timeoutMillis * 2 ** (attempt - 1))` (see the parentheses), which - in case `config.timeout` is `undefined` - would cause the `.min` operator to always resolve to `this.timeoutMillis` (the first expression).

Did you want to use `(config.timeout ?? this.timeoutMillis) * 2 ** (attempt - 1)`? (see parentheses)

src/resource_clients/request_queue.ts
```diff
@@ -39,14 +42,17 @@ export class RequestQueueClient extends ResourceClient {
         });
 
         this.clientKey = userOptions.clientKey;
-        this.timeoutMillis = userOptions.timeoutSecs ? userOptions.timeoutSecs * 1e3 : undefined;
+        this.timeoutMillis = userOptions.timeoutSecs;
```
barjin, line 45: This looks confusing (millis / secs). Perhaps you want to rename the class field? (I see below you're using it as seconds).

src/resource_clients/request_queue.ts
```diff
@@ -79,7 +88,9 @@ export class RequestQueueClient extends ResourceClient {
         const response = await this.httpClient.call({
             url: this._url('head'),
             method: 'GET',
-            timeout: this.timeoutMillis,
+            timeout: this.timeoutMillis
+                ? Math.min(SMALL_TIMEOUT_SECS * 1000, this.timeoutMillis)
```
barjin, line 92: yeah, it's confusing, millis are now seconds, so you'd need to multiply this too :sweat_smile:

Please revisit this :pray:

barjin (changes requested):
+1 for extracting the consts somewhere, so we don't have those in multiple places

src/resource_clients/dataset.ts
```diff
@@ -9,6 +9,10 @@ import type { ApifyRequestConfig, ApifyResponse } from '../http_client';
 import type { PaginatedList } from '../utils';
 import { cast, catchNotFoundOrThrow, pluckData } from '../utils';
 
+const SMALL_TIMEOUT_SECS = 5; // For fast and common actions. Suitable for idempotent actions.
+const MEDIUM_TIMEOUT_SECS = 30; // For actions that may take longer.
+const DEFAULT_TIMEOUT_SECS = 360; // 6 minutes
```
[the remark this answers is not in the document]
janbuchar: the python version also repeats those, but yeah, let's do that

janbuchar:
Sorry for the confusion around secs/millis - I changed my approach when working on this and stopped when the client_timeouts test started to pass :facepalm:

barjin (approved):

src/base/resource_client.ts
```diff
@@ -13,16 +13,21 @@ import { ApiClient } from './api_client';
...
+export const MEDIUM_TIMEOUT_SECS = 30; // For actions that may take longer.
+export const DEFAULT_TIMEOUT_SECS = 360; // 6 minutes
+
 /**
  * Resource client.
  * @private
  */
 export class ResourceClient extends ApiClient {
-    protected async _get<T, R>(options: T = {} as T): Promise<R | undefined> {
+    protected async _get<T, R>(options: T = {} as T, timeoutSecs?: number): Promise<R | undefined> {
```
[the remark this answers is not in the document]
janbuchar: `ApifyClient` and `HttpClient` already accept `timeoutSecs`, so I'm afraid that ship has sailed... But you have a good point that these methods are only used internally, so it would make the internals of the resource clients a bit more uniform.

janbuchar:
@B4nan @barjin I switched to milliseconds almost everywhere and I think the result is in fact much cleaner. If you want to skim it, go ahead. Otherwise I'll merge it later today.

PR #708, opened 2025-06-16
Feature/new acts sorting param sort by

protoss70:
Added new sorting param sortBy for the new feature described in this [Issue](https://app.zenhub.com/workspaces/platform-team-5f6454160d9f82000fa6733f/issues/gh/apify/apify-core/20997)

I will wait until [Apify-Core PR](https://github.com/apify/apify-core/pull/21566) is deployed and then will merge this.

### DIFF
--- a/src/resource_clients/actor_collection.ts
+++ b/src/resource_clients/actor_collection.ts
@@ -28,7 +28,7 @@ export class ActorCollectionClient extends ResourceCollectionClient {
                 limit: ow.optional.number,
                 offset: ow.optional.number,
                 desc: ow.optional.boolean,
-                sortBy: ow.optional.string.oneOf(['createdAt', 'lastRunStartedAt']),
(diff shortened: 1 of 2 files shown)


tobice (approved):

B4nan (approved):
lgtm, just one note about the PR title (which applies to all the repositories, especially libraries), we follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), so it should be in the form `feat: ...` where the `...` will end up in the generated changelog - that part should be imperative and should describe what changed so users can understand it. Only `feat/fix/perf` commit types will end up in the changelog.

For this PR, I would go with something like `feat: allow sorting of Actors collection` or similar (I changed it already but feel free to improve it). The same applies to fixes, the commit type prefix is not part of this "sentence", so while it might look weird at first, using something like `fix: fix this and that` is the correct way, while `fix: this and that` is not.

PR #714, opened 2025-07-02
fix: add `eventData` to `WebhookDispatch` type

valekjo:
The field is returned from the API, but the type was missing it.

In case of `TEST` event, it's null.


katacek (approved):
Nice, thank you! 🚀

vladfrangu (approved):

PR #717, opened 2025-07-14
docs: switch to `@docusaurus/faster`

barjin:
Switching to SWC from `babel` seems to fix the crashing API documentation issue

closes https://github.com/apify/apify-docs/issues/1686


katzino (approved):
😍

PR #720, opened 2025-07-29
feat: add new methods Dataset.getPublicItemsUrl & KeyValueStore.getPublicKeysUrl

danpoletaev:
When storage resources (Datasets or Key-Value Stores) are set to Restricted, accessing or sharing their data externally becomes difficult due to limited permissions. This PR introduces functionality to generate signed URLs that allow controlled external access to these resources without adding token to the request.

This PR introduces methods to generate signed URLs for Dataset items and Key-Value Store records:
1. **Datasets**
`dataset(:datasetId).getPublicItemsUrl(options, expiresInMillis)`
→ Returns a signed URL like:
`/v2/datasets/:datasetId/items?signature=xxx`

2. Key-Value Stores
`keyValueStore(:storeId).getPublicKeysUrl(options, expiresInMillis)`
→ Returns a signed URL like:
`/v2/key-value-stores/:storeId/keys?signature=xxx`

🕒 Expiration:
The `expiresInMillis` parameter defines how long the signature is valid.
- If provided, the URL will expire after the specified time.
- If omitted, the URL will never expire.

Note: The signature is included only if the token has WRITE access to the storage. Otherwise, an unsigned URL is returned.

P.S. We're not yet exposing `urlSigningSecretKey` for datasets, it will be released after [PR](https://github.com/apify/apify-core/pull/22173) is merged.

[More context here](https://www.notion.so/apify/Signed-Dataset-Items-KV-store-record-URLs-224f39950a2280158a6bd82bc2e2ebb5?source=copy_link)

### DIFF
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -12,7 +12,7 @@ import {
 } from '../base/resource_client';
 import type { ApifyRequestConfig, ApifyResponse } from '../http_client';
 import type { PaginatedList } from '../utils';
-import { cast, catchNotFoundOrThrow, pluckData } from '../utils';
+import { applyQueryParamsToUrl, cast, catchNotFoundOrThrow, createStorageSignature, pluckData } from '../utils';
 
 export class DatasetClient<
     Data extends Record<string | number, any> = Record<string | number, unknown>,
@@ -163,6 +163,55 @@ export class DatasetClient<
         return undefined;
     }
 
+    /**
+     * Returns a URL that can be used to access dataset items.
+     *
+     * If the token has permission to access the dataset's URL signing key,
+     * the URL will include a signature to verify its authenticity.
+     *
+     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
+     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * If not provided, the URL will not expire.
+     *
+     * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
+     *
+    */
+    async getPublicItemsUrl(options: DatasetClientListItemOptions = {}, expiresInMillis?: number): Promise<string> {
+        ow(
+            options,
+            ow.object.exactShape({
+                clean: ow.optional.boolean,
+                desc: ow.optional.boolean,
+                flatten: ow.optional.array.ofType(ow.string),
+                fields: ow.optional.array.ofType(ow.string),
+                omit: ow.optional.array.ofType(ow.string),
+                limit: ow.optional.number,
+                offset: ow.optional.number,
+                skipEmpty: ow.optional.boolean,
+                skipHidden: ow.optional.boolean,
+                unwind: ow.optional.any(ow.string, ow.array.ofType(ow.string)),
+                view: ow.optional.string,
+            }),
+        );
+
+        const dataset = await this.get();
+
+        let publicItemsUrl = new URL(this._url('items'));
+        
+        if (dataset?.urlSigningSecretKey) {
+            const signature = createStorageSignature({
+                resourceId: dataset.id,
+                urlSigningSecretKey: dataset.urlSigningSecretKey,
+                expiresInMillis,
+            });
+            publicItemsUrl.searchParams.set('signature', signature);
+        }
+
+        publicItemsUrl = applyQueryParamsToUrl(publicItemsUrl, options);
+
+        return publicItemsUrl.toString();
+    }
+
     private _createPaginationList(response: ApifyResponse, userProvidedDesc: boolean): PaginatedList<Data> {
         return {
             items: response.data,
@@ -191,6 +240,7 @@ export interface Dataset {
     stats: DatasetStats;
     fields: string[];
     generalAccess?: STORAGE_GENERAL_ACCESS | null;
+    urlSigningSecretKey?: string | null;
 }
 
 export interface DatasetStats {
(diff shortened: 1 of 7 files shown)


src/resource_clients/key_value_store.ts
```diff
@@ -75,6 +85,48 @@ export class KeyValueStoreClient extends ResourceClient {
...
+     * the URL will include a signature to verify its authenticity.
+     *
+     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
+     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * If not provided, the URL will not expire.
+     *
+     * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
+     *
+     */
+    async getPublicKeysUrl(options: KeyValueClientListKeysOptions = {}, expiresInMillis?: number) {
```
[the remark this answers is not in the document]
danpoletaev: Sure, we can change the name. How about `getPublicKeyListUrl`?

src/resource_clients/dataset.ts
```diff
@@ -163,6 +163,55 @@ export class DatasetClient<
         return undefined;
     }
 
+    /**
+     * Returns a URL that can be used to access dataset items.
+     *
+     * If the token has permission to access the dataset's URL signing key,
```
tobice, line 169:
```suggestion
     * If the client has permission to access the dataset's URL signing key,
```

(nit) Maybe like this? I'd find this more intuitive from the perspective a client user.

src/resource_clients/dataset.ts
```diff
@@ -191,6 +240,7 @@ export interface Dataset {
     stats: DatasetStats;
     fields: string[];
     generalAccess?: STORAGE_GENERAL_ACCESS | null;
+    urlSigningSecretKey?: string | null;
```
tobice, line 243: We're also now returning `publicItemsUrl` in the dataset response, right? Maybe let's add it to the type?

src/resource_clients/dataset.ts
```diff
@@ -163,6 +163,55 @@ export class DatasetClient<
...
+     * the URL will include a signature to verify its authenticity.
+     *
+     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
+     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * If not provided, the URL will not expire.
+     *
+     * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
+     *
+     */
+    async getPublicItemsUrl(options: DatasetClientListItemOptions = {}, expiresInMillis?: number): Promise<string> {
```
tobice, line 179:
```suggestion
    async createPublicItemsUrl(options: DatasetClientListItemOptions = {}, expiresInMillis?: number): Promise<string> {
```

Maybe call it _create_ to make it explicit that the URL is actually being crafted on the go? And also to distinguish it from the already provided `publicItemsUrl`?

I'd still wouldn't be surprised if users were a bit confused by this 😄

src/resource_clients/key_value_store.ts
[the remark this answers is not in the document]
tobice: Maybe we could do `getKeysPublicUrl`? 

I was thinking about it already on the API PR... to me it makes more sense language-wise. It's not URL for public keys, it's a public URL for keys. But AI disagreed with my thinking so I let it be 😄

src/utils.ts
```diff
@@ -260,3 +262,45 @@ export function asArray<T>(value: T | T[]): T[] {
...
+    urlSigningSecretKey,
+    expiresInMillis,
+    version = 0,
+}: {
+    resourceId: string;
+    urlSigningSecretKey: string;
+    expiresInMillis: number | undefined;
+    version?: number;
+}) {
+    const expiresAt = expiresInMillis ? new Date().getTime() + expiresInMillis : null;
```
tobice, line 283:
```suggestion
    const expiresAt = expiresInMillis ? new Date().getTime() + expiresInMillis : '0';
```

(opt) Maybe this? 

Then in API you'll be able to do:

```typescript
const parsedExpiresAtMillis = parseInt(expiresAtMillis, 10);
if (parsedExpiresAtMillis > 0 && parsedExpiresAtMillis < new Date().getTime()) {
    // signature expired
    return false;
}

**Review by tobice — CHANGES_REQUESTED** [review]
Generally makes sense but I left a few nits and would like to take another look once it is finished.

File: src/utils.ts (hunk 5 above)
Comment (danpoletaev) [reply src/utils.ts:283]: Honestly, I'd keep it as `null, but I updated comments in API 👇 

```
// For non-expiring signatures, we pass 'null' as timestamp.
// If the timestamp is 'null', we don't need to check for expiration.
if (expiresAtMillis !== 'null' && parseInt(expiresAtMillis, 10) < new Date().getTime()) {
  // signature expired
  return false;
}
```

tobice (approved):

janbuchar (commented):
Please update the PR title and description to match the new name 🙂 After that, feel free to merge

src/utils.ts
danpoletaev: well, but it is not required to check for `expiresAtMillis !== 'null'`, I thought it would be more self-explanatory 😄 

`NaN < new Date().getTime()` would always return false

danpoletaev: Okay, you're right 😄
 
In python it will looks ugly, since python has `None` not `null`.

`expires_at_str = str(expires_at) if expires_at is not None else 'null'`

Changing it to 0

PR #724, opened 2025-08-05
fix: KV store createKeysPublicUrl wrong URL

danpoletaev:
After release of new methods `KeyValueStore.createKeysPublicUrl` and `Dataset.createItemsPublicUrl` ([PR here](https://github.com/apify/apify-client-js/pull/720)),  I noticed that in `KeyValueStore.createKeysPublicUrl` I used incorrect path `items` instead of `keys`.

This PR fixes URL path and improves tests.


tobice (approved):
Nice catch 👍 

Ideally we update the integration tests to use the JS client. Then we'll have a nice, true e2e test 😊

PR #725, opened 2025-08-06
feat: add keyValueStore.getRecordPublicUrl

danpoletaev:
This PR adds new method to KV store client:
- `getRecordPublicUrl(key: string)`

Note: We already have this same method ([KeyValueStore.getPublicUrl](https://github.com/apify/apify-sdk-js/blob/master/packages/apify/src/key_value_store.ts#L19)) in SDK.

### DIFF
--- a/src/resource_clients/key_value_store.ts
+++ b/src/resource_clients/key_value_store.ts
@@ -110,7 +110,7 @@ export class KeyValueStoreClient extends ResourceClient {
      * Generates a URL that can be used to access key-value store keys.
      *
      * If the client has permission to access the key-value store's URL signing key,
-     * the URL will include a signature to verify its authenticity.
+     * the URL will include a signature which will allow the link to work even without authentication.
      *
      * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
      * This value sets the expiration duration in milliseconds from the time the URL is generated.
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -168,7 +168,7 @@ export class DatasetClient<
      * Generates a URL that can be used to access dataset items.
      *
      * If the client has permission to access the dataset's URL signing key,
-     * the URL will include a signature to verify its authenticity.
+     * the URL will include a signature which will allow the link to work even without authentication.
      *
      * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
      * This value sets the expiration duration in milliseconds from the time the URL is generated.


janbuchar:
Pardon me being blunt, but what's the endgame here? Are users expected to stop using `KeyValueStore.getPublicUrl` and use this instead? Should both co-exist in harmony?

danpoletaev:
> Pardon me being blunt, but what's the endgame here? Are users expected to stop using `KeyValueStore.getPublicUrl` and use this instead? Should both co-exist in harmony?

`KeyValueStore.getPublicUrl` is SDK method, now I've added same logic to the client. While creating new Actors, people would use mainly SDK, but there would be other cases when you'd want to use it without SDK 🙃

janbuchar:
> > Pardon me being blunt, but what's the endgame here? Are users expected to stop using `KeyValueStore.getPublicUrl` and use this instead? Should both co-exist in harmony?
> 
> `KeyValueStore.getPublicUrl` is SDK method, now I've added same logic to the client. While creating new Actors, people would use mainly SDK, but there would be other cases when you'd want to use it without SDK 🙃

Huh, okay, what would those cases be? From the perspective of an Actor developer (or anyone primarily using the SDK), having two slightly different (or are they?) ways to achieve the same thing is very confusing.

danpoletaev:
> > > Pardon me being blunt, but what's the endgame here? Are users expected to stop using `KeyValueStore.getPublicUrl` and use this instead? Should both co-exist in harmony?
> > 
> > 
> > `KeyValueStore.getPublicUrl` is SDK method, now I've added same logic to the client. While creating new Actors, people would use mainly SDK, but there would be other cases when you'd want to use it without SDK 🙃
> 
> Huh, okay, what would those cases be? From the perspective of an Actor developer (or anyone primarily using the SDK), having two slightly different (or are they?) ways to achieve the same thing is very confusing.

For example you just want to use our API, why would you download whole Apify SDK with Crawlee for that?
I suppose our integration team will mostly use client instead of SDK.

Also, in `apify-cli` I need this method to sign a tarball ZIP of the build.

Methods are nearly identical in SDK and client

src/resource_clients/key_value_store.ts
```diff
@@ -85,6 +85,27 @@ export class KeyValueStoreClient extends ResourceClient {
         return cast(parseDateFields(pluckData(response.data)));
     }
 
+    /**
+     * Generates a URL that can be used to access key-value store record.
+     *
+     * If the client has permission to access the key-value store's URL signing key,
+     * the URL will include a signature to verify its authenticity.
```
tobice, line 92:
```suggestion
     * the URL will include a signature which will allow the link to work even without authentication.
```

tobice (approved):
Approving. Agreed with the points here... the end game should be that we can just proxy the client implementation from SDK.

And perhaps the naming should be unified 😄

janbuchar:
> And perhaps the naming should be unified 😄

Yes please!

PR #733, opened 2025-08-26
refactor!: rename expiresInMillis to expiresInSecs in create storage content URL

danpoletaev:
This PR changes `expiresInMillis` to `expiresInSecs` and moves it to options in create storage content URLs:
- `KeyValueStore.createKeysPublicUrl(options)`
- `Dataset.createItemsPublicUrl(options)`

Note: I'm aware that this can potentially be breaking change, but we suppose that no one started using it yet.

More context: https://apify.slack.com/archives/C01VBUV81UZ/p1756111935325179

### DIFF
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -170,13 +170,13 @@ export class DatasetClient<
      * If the client has permission to access the dataset's URL signing key,
      * the URL will include a signature which will allow the link to work even without authentication.
      *
-     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
-     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * You can optionally control how long the signed URL should be valid using the `expiresInSecs` option.
+     * This value sets the expiration duration in seconds from the time the URL is generated.
      * If not provided, the URL will not expire.
      *
      * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
      */
-    async createItemsPublicUrl(options: DatasetClientListItemOptions = {}, expiresInMillis?: number): Promise<string> {
+    async createItemsPublicUrl(options: DatasetClientListItemOptions = {}, expiresInSecs?: number): Promise<string> {
         ow(
             options,
             ow.object.exactShape({
@@ -202,7 +202,7 @@ export class DatasetClient<
             const signature = createStorageContentSignature({
                 resourceId: dataset.id,
                 urlSigningSecretKey: dataset.urlSigningSecretKey,
-                expiresInMillis,
+                expiresInMillis: expiresInSecs ? expiresInSecs * 1000 : undefined,
             });
             createdItemsPublicUrl.searchParams.set('signature', signature);
         }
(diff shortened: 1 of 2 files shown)


src/resource_clients/dataset.ts
```diff
@@ -170,13 +170,13 @@ export class DatasetClient<
...
-     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
-     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * You can optionally control how long the signed URL should be valid using the `expiresInSecs` option.
+     * This value sets the expiration duration in seconds from the time the URL is generated.
      * If not provided, the URL will not expire.
      *
      * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
      */
-    async createItemsPublicUrl(options: DatasetClientListItemOptions = {}, expiresInMillis?: number): Promise<string> {
+    async createItemsPublicUrl(options: DatasetClientListItemOptions = {}, expiresInSecs?: number): Promise<string> {
```
janbuchar, line 179: While you're refactoring, are you sure you don't want to add `expiresInSecs` to `options` so that the caller has to write `expiresInSecs: 30`, which makes the unit of the number obvious?

B4nan: What about putting this to the `DatasetClientListItemOptions` (or extending that, adding the new option)?

That way we could just deprecate the second parameter, leaving it in milliseconds.

danpoletaev: Could be. Currently `DatasetClientListItemOptions` is used by multiple methods: `listItems()` and `createItemsPublicUrl()`. 

You think DX will be better with `expiresInSecs`? My intention is that maybe developers won't even notice, that they can create expiring signatures, but I think description should be enough, so I am ok with that

src/resource_clients/key_value_store.ts
```diff
@@ -112,20 +112,21 @@ export class KeyValueStoreClient extends ResourceClient {
...
-     * You can optionally control how long the signed URL should be valid using the `expiresInMillis` option.
-     * This value sets the expiration duration in milliseconds from the time the URL is generated.
+     * You can optionally control how long the signed URL should be valid using the `expiresInSecs` option.
+     * This value sets the expiration duration in seconds from the time the URL is generated.
      * If not provided, the URL will not expire.
      *
      * Any other options (like `limit` or `prefix`) will be included as query parameters in the URL.
      */
-    async createKeysPublicUrl(options: KeyValueClientListKeysOptions = {}, expiresInMillis?: number) {
+    async createKeysPublicUrl(options: KeyValueClientCreateKeysUrlOptions = {}) {
```
B4nan, line 121: I would rather keep the second argument as it is, and add a `log.deprecated` to it when the value is present, just in case it's already used somewhere. I can imagine it will be mostly testing actors from your end, but still, even for those, it feels like a better thing to have a proper warning than just a compilation failure.

We will be releasing major bumps for all the packages relatively soon (let's say by the end of this year), so we can remove it there.

tobice (approved):

PR #736, opened 2025-08-27
docs: version only major bumps

B4nan:
Only major versions are now producing docs snapshots. On release, old version is removed and generated from scratch, so it always reflects the latest stable one.

image

I am sure this could be done in another way, feels a bit hacky, but it works 🙃 

Closes https://github.com/apify/apify-docs/issues/1837


.github/workflows/release.yaml
```diff
@@ -121,6 +121,58 @@ jobs:
...
+                  yarn
+
+            - name: Snapshot the current version
+              run: |
+                  cd website
+                  VERSION=$(jq -r '.version' ../package.json)
+                  MAJOR=$(echo $VERSION | cut -d. -f1)
+                  rm -rf versioned_docs/version-$MAJOR
+                  rm -rf versioned_sidebars/version-$MAJOR-sidebars.json
+                  jq --arg major "$MAJOR" 'map(select(. != $major))' version.json > tmp.json && mv tmp.json version.json
```
vladfrangu, line 164: This magic incantation.... I love it, and I'm sure it works perfectly fine, but wow is this ever smth that we could've probably done in a nodejs script instead 😄

vladfrangu (approved):
At a glance this looks fine, maybe get another pair of eyes from @janbuchar too? :D

barjin (approved):
lgtm (beside the my comments), let's see how it works in the wild :+1:

PR #743, opened 2025-09-05
feat: add forcePermissionLevel run option

tobice:
We want Actor developers to be able to easily test their full permission Actors with limited permissions without changing the Actor configuration for everybody or redeploying the Actor under a different name.  For that reason, we have introduced a new `forcePermissionLevel` run option. This PR adds the option to the client.

Full context here: https://github.com/apify/apify-core/pull/22681

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -2,7 +2,7 @@ import type { AxiosRequestConfig } from 'axios';
 import ow from 'ow';
 
 import type { RUN_GENERAL_ACCESS } from '@apify/consts';
-import { ACT_JOB_STATUSES, META_ORIGINS } from '@apify/consts';
+import { ACT_JOB_STATUSES, ACTOR_PERMISSION_LEVEL , META_ORIGINS } from '@apify/consts';
 
 import type { ApiClientSubResourceOptions } from '../base/api_client';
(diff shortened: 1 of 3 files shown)


stepskop (approved):

src/resource_clients/actor.ts
```diff
@@ -2,7 +2,7 @@ import type { AxiosRequestConfig } from 'axios';
 import ow from 'ow';
 
 import type { RUN_GENERAL_ACCESS } from '@apify/consts';
-import { ACT_JOB_STATUSES, META_ORIGINS } from '@apify/consts';
+import { ACT_JOB_STATUSES, ACTOR_PERMISSION_LEVEL, META_ORIGINS } from '@apify/consts';
```
B4nan, line 5: looks like the `ACTOR_PERMISSION_LEVEL` was introduced in `@apify/consts` v2.42.0, but we depend on ^v2.25.0 here, we should bump the lower bound to be sure

https://github.com/apify/apify-client-js/blob/master/package.json#L65

B4nan (changes requested):

src/resource_clients/actor.ts
tobice: Bumped :+1:

PR #760, opened 2025-10-02
feat: move restartOnError from Actor to Run options

DaveHanns:
Changes related to move of `restartOnError` from Actor's top level to its run options.
 
Core: https://github.com/apify/apify-core/pull/23381
Worker: https://github.com/apify/apify-worker/pull/1515
Shared: https://github.com/apify/apify-shared-js/pull/546
Py Client: https://github.com/apify/apify-client-python/pull/508
Docs: https://github.com/apify/apify-docs/pull/1965

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -75,7 +75,16 @@ export class ActorClient extends ResourceClient {
             }),
         );
 
-        const { waitForFinish, timeout, memory, build, maxItems, maxTotalChargeUsd, restartOnError, forcePermissionLevel } = options;
+        const {
+            waitForFinish,
+            timeout,
+            memory,
+            build,
+            maxItems,
+            maxTotalChargeUsd,
+            restartOnError,
+            forcePermissionLevel,
+        } = options;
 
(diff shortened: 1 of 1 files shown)


B4nan:
> That is BC, but it should be reflected.

Not sure what you mean by that. If it's a BC, it should be reflected in a major bump of the library version.

Let's just mark it as deprecated, pointing to the correct place where it should be used.

DaveHanns:
> > That is BC, but it should be reflected.
> 
> Not sure what you mean by that. If it's a BC, it should be reflected in a major bump of the library version.
> 
> Let's just mark it as deprecated, pointing to the correct place where it should be used.

Sorry for the confusion. What I meant is that leaving the flags at the top level of Actor create and update payload types is ok, as we will still accept the flags there for a while anyway.

`restartOnError` at top-level of `Actor` can be kept there and flagged deprecated, but it will never appear in the actual data returned to the user as after the migration, it will be permanently moved to the `defaultRunOptions`.

B4nan (approved):

Issue #753, opened 2025-09-17
Make `apify-client` runtime-independent


barjin:
Because of importing `node:xyz` packages, the Apify shared packages (e.g. `@apify/utilities`) and the client require polyfills when bundled in a web application, or used in a non-node server-side runtime. This bloats the bundle size and complicates initial project setup (including Apify Console itself).

All(?) of the used behaviour now has Web API alternatives, which tend to be implemented across all major runtimes and browsers. Swapping the current `node` imports for the native implementations might solve this issue.

Needs investigation.

Solves #183 
Solves #564 
Related to https://github.com/apify/apify-shared-js/issues/537

vdusek:
resolved by https://github.com/apify/apify-client-js/pull/1052

PR #767, opened 2025-10-20
perf: dynamic imports for certain `node:` packages

barjin:
Changes some `node:`-scoped imports into dynamic imports to remove the need for polyfilling / bundling those packages (which are unused in the browser environment anyway).

Related to #753 

Note that the issue is not fully solved because of the dependencies from https://github.com/apify/apify-shared-js and the calls to `node:os` in `HttpClient` (see below), which would require bigger refactors (ESM imports are asynchronous, so cannot be done in a constructor).

https://github.com/apify/apify-client-js/blob/a9c65ec2ed75e1141c33d5d7ad0b1dec7fc0654c/src/http_client.ts#L109

### DIFF
--- a/rsbuild.config.ts
+++ b/rsbuild.config.ts
@@ -31,7 +31,7 @@ export default defineConfig({
         externals: {
             'node:util': 'node:util',
             'node:zlib': 'node:zlib',
-        }
+        },
     },
     tools: {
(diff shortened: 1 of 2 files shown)


B4nan:
sidenote: you need to be a bit more careful about the wording :]

image

PR #768, opened 2025-10-21
ci: Set up OIDC NPM publishing (initial attempt)

### LINKED ISSUE #732: Set up OIDC npm publishing
https://github.blog/changelog/2025-07-31-npm-trusted-publishing-with-oidc-is-generally-available/

janbuchar:
- closes #732

This is the first one in a series of PRs to set up OIDC based publishing to npm. It deserves special attention because it will shape the way how we approach this for the rest of the packages.

## The Challenge

npm only allows configuring one workflow as the trusted publisher. That's it. But our repositories currently use two workflow files, one for pre-releases and one for stable ones.

## Selected solution

I simply merged the two workflows in a single file, almost as if they were still separate. Both branches depend on a `determine_path` job that analyzes the event that triggered the workflow. From that point, the execution paths are completely disjoint.

This has a major advantage - both the review of the changes and of the workflow are fairly straightforward. Experience suggests that attempting to "DRY this up" makes the resulting workflow shorter, but very hard to follow.

## Alternatives

1. the DRY approach where we try to branch as deep in the workflow as possible
    - not very practical (see reasoning above)
2. making a separate workflow just for NPM publishing and triggering it via Github API (because calling a reusable workflow won't pass the npm trusted publisher check)
    - this would require some non-trivial scripting, but it seems feasible
    - I'm unsure what the resulting workflow visualization would look like - it could be a mess
    - but we could keep the two separate workflows
    - see #770
3. abandon this and automate token rotation instead :see_no_evil: 
4. publish to npm in a separate workflow triggered each time a tag is pushed (we'd have to start making beta tags again)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,7 +28,7 @@ on:
             - '**' # Ignore all tags to prevent duplicate builds when tags are pushed.
 
 permissions:
-    id-token: write  # Required for OIDC
+    id-token: write # Required for OIDC
     contents: read
 
 concurrency:


vladfrangu:
Random question while I am omw home to review proper: couldn't we use composite actions for the release flows? That is, we define the steps as composite actions and then based on input run the pre release or non pre release flow 👀

Not sure if it even works but 🤷‍♂️ thought I'd throw that into the bowl

.github/workflows/release.yaml
```diff
@@ -20,14 +20,54 @@ on:
...
+        steps:
+            - name: Decide path
+              id: decide
+              run: |
+                  if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+                      echo "path=stable" >> $GITHUB_OUTPUT
+                  elif [[ "${{ github.event_name }}" == "push" ]] && \
+                       [[ ! "${{ github.event.head_commit.message }}" =~ ^docs ]] && \
+                       [[ ! "${{ github.event.head_commit.message }}" =~ ^ci ]] && \
+                       [[ "${{ github.repository }}" =~ ^apify/ ]]; then
```
vladfrangu: I would move the checks for repository out into each job instead

B4nan (approved):
looks good to me, not great, not terrible. i'll surely prefer readability over deduplication. 

sidenote: i am quite sure that after we finish migration in the last repo, they will allow whitelisting multiple workflows :trollface:

PR #788, opened 2025-11-19
feat: Replace `agentkeepalive` with native Node.js HTTP agents for proxy support

tducret:
Hello 👋 

Here is a PR proposition to be able to use `apify` CLI in a sandboxed environment (in my case, Claude code in the browser).

At the moment, we can't do `apify login` or `apify info` in such environment, because the **apify-client** doesn't handle the proxy configuration properly.
The commands are stuck.

Digging into it, I found out that the `agentkeepalive` module doesn't support proxy (see https://github.com/node-modules/agentkeepalive/issues/22#issuecomment-661055269 for instance).

I first attempted to use [proxy-agents](https://github.com/TooTallNate/proxy-agents), but it turns out the built-in agents can support both keepalive and proxies out of the box.

So this PR removes `agentkeepalive` while maintaining the same keep-alive functionality and adding proxy support capability.

### Test
To test it locally, I created a transparent proxy (we don't intercept requests to avoid self-certificate issues):
```sh
mitmdump --listen-port 8000 --ignore-hosts '.*:*' --set dns_log=true
```

And in another terminal, I tested the Apify client with :
```sh
export HTTPS_PROXY=http://localhost:8000
export APIFY_TOKEN=[REPLACE_WITH_VALID_TOKEN]

node get-user-info.js  # See code below
```

```js
#!/usr/bin/env node

const { ApifyClient } = require('./dist/index.js');

async function main() {
    // Get token from environment variable
    const token = process.env.APIFY_TOKEN;

    if (!token) {
        console.error('Error: APIFY_TOKEN environment variable is not set');
        process.exit(1);
    }

    try {
        const client = new ApifyClient({ token });

        console.log('Fetching user information...\n');
        const userInfo = await client.user().get();

        console.log('User Information:');
        console.log(JSON.stringify(userInfo, null, 2));
    } catch (error) {
        console.error('Error fetching user information:');
        console.error(error.message);
        if (error.statusCode) {
            console.error(`Status Code: ${error.statusCode}`);
        }
        process.exit(1);
    }
}

main();
```

### DIFF
--- a/rsbuild.config.ts
+++ b/rsbuild.config.ts
@@ -31,6 +31,8 @@ export default defineConfig({
         externals: {
             'node:util': 'node:util',
             'node:zlib': 'node:zlib',
+            'node:http': 'node:http',
+            'node:https': 'node:https',
             crypto: 'node:crypto',
         },
     },


janbuchar (approved):
I think this is good to go, but let's give @barjin a chance to review it too. Thanks @tducret!

PR #791, opened 2025-11-20
feat: Add CONNECT tunneling support for HTTP proxy in sandboxed environments

tducret:
This PR is a follow-up of PR #788, which unfortunately didn't fully fix the use of **apify-cli** in Claude Code sandbox (in the browser, not the local one).

Currently, this is what's happening :

1. Proxy receives the request
2. Proxy returns: HTTP/1.1 301 Moved Permanently
3. Location: https://api.apify.com:443/v2/users/me (adds explicit :443 port)
4. axios follows redirect to https://api.apify.com:443/v2/users/me
5. Proxy strips the :443 and redirects back to https://api.apify.com/v2/users/me
6. **Infinite loop!**

This is a known `axios` problem, referenced in several **unsolved** issues :
- https://github.com/axios/axios/issues/6725
- https://github.com/axios/axios/issues/2072
- https://github.com/axios/axios/issues/925
- https://github.com/axios/axios/issues/4531

To fix this issue, this PR adds support for CONNECT tunneling through HTTP/HTTPS proxies, using [proxy-agents](https://github.com/TooTallNate/proxy-agents) modules.

What this does :
1. HttpsProxyAgent sends CONNECT api.apify.com:443
2. Proxy establishes tunnel (returns 200 OK)
3. Client does TLS handshake through tunnel
4. Encrypted traffic flows through tunnel
5. Proxy cannot see/modify the HTTPS traffic (acts like classic proxy)
6. **No redirects, no loops!**

~By precaution, we keep here the current http[s] agent implementation when the `HTTP[S]_PROXY` env vars are not defined.~ (Changed since commit 89571f926747e0e23cdb1bf1a63ba1a4bb480f3f)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,9 +17,8 @@
                 "async-retry": "^1.3.3",
                 "axios": "^1.6.7",
                 "content-type": "^1.0.5",
-                "http-proxy-agent": "^7.0.2",
-                "https-proxy-agent": "^7.0.6",
                 "ow": "^0.28.2",
+                "proxy-agent": "^6.5.0",
                 "tslib": "^2.5.0",
                 "type-fest": "^4.0.0"
             },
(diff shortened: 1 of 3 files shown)


tducret:
Hey @janbuchar @B4nan, the PR #788 was not enough to be able to use the **Apify CLI** in Claude code sandbox 😢 
Fortunately, I have prepared this PR to fix it.
Could you please check it out?

janbuchar (approved):
Thanks @tducret!

Issue #632, opened 2025-01-28
`.actor().call()` method to set the correct timeout, show the progress in status message, and stream logs


mtrunkat:
I was trying out https://apify.com/jakub.kopecky/llmstxt-generator Actor, the experience was not great because of the following:

### Timeout

The Actor above was started with a timeout of 18,000 seconds, but the WCC is triggered with a default timeout of 360,000. So, it may happen that the original Actor timeouts, but the WCC will continue running. IMHO, in this case, we should set the timeout for the remaining time for the original Actor.

There might be cases when this is not appropriate, so this behavior could be opt-in or out.

### Logs

It's called WCC underneath, which may take a long time to finish in the case of a large website. This means that the Actor seem to get stuck on the following log:
```
2025-01-23T13:56:06.535Z ACTOR: Pulling Docker image of build OQWIcf5rmeLt4icyd from repository.
2025-01-23T13:56:08.308Z ACTOR: Creating Docker container.
2025-01-23T13:56:08.850Z ACTOR: Starting Docker container.
2025-01-23T13:56:11.052Z [apify] INFO  Initializing Actor...
2025-01-23T13:56:11.054Z [apify] INFO  System info ({"apify_sdk_version": "2.1.0", "apify_client_version": "1.8.1", "crawlee_version": "0.4.5", "python_version": "3.12.8", "os": "linux"})
2025-01-23T13:56:11.119Z [apify] INFO  Starting the "apify/website-content-crawler" actor for URL: https://docs.apify.com/
```

So, I am thinking about improving the `.actor().call()` method in SDK/client the way that it enables developers to optionally stream the log from the Actor called via a `.call()` to provide progress/context info.

### Status message

Finally, it displays a dummy status message that does not communicate progress. The call could automatically update the status message, for example, here, with:

> Running Website Content Crawler: processed 235/7876

mtrunkat:
You can see @MQ37 improving this on the Actor side: https://github.com/apify/actor-llmstxt-generator/pull/10

Pijukatel:
Regarding the timeouts. Currently there is no public way of getting the runtime of the actor run. Specifically runtime of last actor run segment, which in the context of this issue is the actor run.  (Resurrected actor run consists of several run segments and from the point of view of API, the actor run is sum of all it's run segments)

[...]

PR #793, opened 2025-11-21
feat: Add `StatusMessageWatcher`

Pijukatel:
### Description

Add the option to log status and status messages of another Actor run. If such Actor run is also redirecting logs from another Actor run, then those logs can propagate all the way to the top through the `StreamedLog` - deep redirection. If the Actor run from which we are redirecting status messages is not redirecting logs from its children, then the log and status redirection remains shallow - only from that Actor run. 

#### Actor.call - default
This will redirect logs by default. Example default redirected line:

> 2025-10-22T13:06:39.476Z redirect-log-tester runId:SoK1VRJxG61tVrgNz -> 2025-10-22T13:06:06.686Z ACTOR: Status: FAILED, Message: Different status message

```javascript
...
await client.actor('redirect-actor-id').call();
```
#### Actor.call - custom Log
This will redirect logs using custom log and logger. Example default redirected line:

> 2025-10-22T13:06:39.476Z customPrefix 2025-10-22T13:06:06.686Z ACTOR: Status: FAILED, Message: Different status message

```javascript
...
await client.actor('redirect-actor-id').call(someInputs, {
    log: new Log({ level: LEVELS.DEBUG, prefix: 'customPrefix', logger: new LoggerActorRedirect()
    }),
```
#### Actor.call - no log redirection

This will disable all log redirection (same as current behavior)

```javascript
...
await client.actor('redirect-actor-id').call(someInputs, {log: null}),
```

#### Actor.run - attaching to already running actor and redirecting statuses and status messages
A typical use case is redirecting statuses and status messages from an Actor that runs in standby.

```javascript
...
const statusMessageWatcher = await client.run('redirect-run-id').getStatusMessageWatcher();

statusMessageWatcher.start();
// Do some stuff while also redirecting statuses from another Actor run
await statusMessageWatcher.stop();
```

#### Example actor with recursive redirection:
https://console.apify.com/actors/IcfIKTIvbmujMmovj/source

### Issues

- Partially implements: #632
- Parity PR in Python: https://github.com/apify/apify-client-python/pull/407


Pijukatel:
Example log from nested actor runs and their redirection:

image

src/resource_clients/run.ts
```diff
@@ -279,24 +279,49 @@ export class RunClient extends ResourceClient {
...
+        const runData = await this.get();
+        const runId = runData ? `${runData.id ?? ''}` : '';
+
+        const actorId = runData?.actId ?? '';
+        const actorData = (await this.apifyClient.actor(actorId).get()) || { name: '' };
+
+        const actorName = runData ? (actorData.name ?? '') : '';
+        const name = [actorName, `runId:${runId}`].filter(Boolean).join(' ');
+
+        return new Log({ level: LEVELS.DEBUG, prefix: `${name} -> `, logger: new LoggerActorRedirect() });
```
barjin:
```suggestion
        const runData = await this.get();
        const runId = runData?.id ?? '';

        const actorId = runData?.actId ?? '';
        const actorData = (await this.apifyClient.actor(actorId).get()) || { name: '' };

        const actorName = actorData?.name ?? '';
        const name = [actorName, `runId:${runId}`].filter(Boolean).join(' ');

        return new Log({ level: LEVELS.DEBUG, prefix: `${name} -> `, logger: new LoggerActorRedirect() });
```

This is the removed code from L282, maybe this PR was made before merging the final version of log streaming?

barjin (commented):
Thank you @Pijukatel! I know it's already in Python Client, and there's an issue asking for this, but still - I'm not completely convinced we need this feature in the Client.

I have a few opinions, in no particular order:

1. The status message is a "discrete" piece of data (unlike e.g. `log`), so polling the API should be the user's job. 
2. (If we still do this), we should allow for consuming the changed status without logging it (e.g. `run().onStatus((status) => { user code })`)
3. (If we don't do this), could this be an option for the logger, e.g. `.getStreamedLog({ statusMessage: true })`? Having a separate method for this doesn't make much sense to me.

cc @janbuchar @B4nan wdyt?

Anyway, regarding the code, a few comments from a quick look:

Pijukatel:
> Thank you @Pijukatel! I know it's already in Python Client, and there's an issue asking for this, but still - I'm not completely convinced we need this feature in the Client.

As far as I am concerned, this is the least useful feature from the [3 new requested features](https://github.com/apify/apify-client-js/issues/632). The other two are already merged. We can wait to see if there is demand for this one or not.

Pijukatel:
Not needed at the moment

Issue #777, opened 2025-11-04
Add better support for pagination


jancurn:
To paginate long lists, users need to implement code like this, which is quite painful:

```js
// Fetch all Actors
let actors = [];
const fetchNextChunk = async (offset = 0) => {
    const limit = 1000;
    const value = await apifyClient.store({ search }).list({ offset, limit });

    if (value.count === 0) {
        return;
    }

    actors.push(...value.items);

    if (value.total > offset + value.count) {
        return fetchNextChunk(offset + value.count);
    }
};
await fetchNextChunk();
```

Can we add some of this to the library, to make it easier to do?

Or maybe there's some better way to do it, but the docs doesn't provide much more info: https://docs.apify.com/api/client/js/reference/interface/PaginatedList

PR #803, opened 2025-12-02
feat: Make storage clients list methods return value also be asyncIterator of relevant data

Pijukatel:
### Description

- Extends the return value of storage list methods by `asyncIterator` that can be used to iterate over individual items:
  - `DatasetClient.listItems`
  - `KeyValueStoreClient.listKeys`
  - `RequestQueueClient.listRequests`

- Generalize `asyncIterator` pagination tests so that each method is tested with options it supports
- Add stricter runtime checks on `limit` and `offset` (They would cause API errors anyway)
- Add a comment about the undocumented behavior of `exclusiveStartId`
- Update example code in docs

### Example usage

It can still be used the same way as before, and additionally, it can be used like this now:

```ts
...
for await (const item of datasetClient.listItems({ limit, offset, chunkSize })) {
    allItems.push(item);
}

console.log(`Overall fetched ${allItems.length} items`);
```

### Issues

- Closes: #777

### DIFF
--- a/src/resource_clients/dataset.ts
+++ b/src/resource_clients/dataset.ts
@@ -153,7 +153,7 @@ export class DatasetClient<
             }),
         );
 
-        const listItems = async (
+        const fetchItems = async (
             datasetListOptions: DatasetClientListItemOptions = {},
         ): Promise<PaginatedList<Data>> => {
             const response = await this.httpClient.call({
@@ -166,7 +166,7 @@ export class DatasetClient<
             return this._createPaginationList(response, datasetListOptions.desc ?? false);
         };
 
-        return this._listPaginatedFromCallback(listItems.bind(this), options);
+        return this._listPaginatedFromCallback(fetchItems, options);
     }
 
     /**
--- a/test/pagination.test.js
+++ b/test/pagination.test.js
@@ -118,11 +118,11 @@ describe('Collection clients list method as async iterable', () => {
 
                 const totalItems = request.params.unnamed ? normalItems + extraItems : normalItems;
 
-                const offset = request.params.offset ? request.params.offset : 0;
-                const limit = request.params.limit ? request.params.limit : 0;
-                const desc = request.params.desc ? request.params.desc : false;
+                const offset = request.params.offset ?? 0;
+                const limit = request.params.limit ?? 0;
+                const desc = request.params.desc ?? false;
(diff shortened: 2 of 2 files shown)


docs/index.md
```diff
@@ -256,27 +256,17 @@ const client = new ApifyClient({ token: 'MY-APIFY-TOKEN' });
...
-    allItems.push(...items);
-
-    // If there are no more items to fetch, exit the loading
-    if (offset + limit >= total) {
-        break;
-    }
-
-    offset += limit;
+for await (const item of datasetClient.listItems({ limit, offset, chunkSize })) {
+    allItems.push(item);
```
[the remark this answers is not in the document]
Pijukatel: The point is only to highlight how the individual items can be easily accessed within the loop. Would this be ok for you?
```
...
for await (const item of datasetClient.listItems({ limit, offset, chunkSize })) {
    // Process individual item
    console.log(item)
...
```

src/resource_clients/dataset.ts
```diff
@@ -73,14 +74,20 @@ export class DatasetClient<
             }),
         );
 
-        const response = await this.httpClient.call({
-            url: this._url('items'),
-            method: 'GET',
-            params: this._params(options),
-            timeout: DEFAULT_TIMEOUT_MILLIS,
-        });
+        const listItems = async (
```
barjin: nit: can we rename this so it doesn't clash with the enclosing method? maybe `fetchItems` ?

src/resource_clients/dataset.ts
```diff
@@ -73,14 +74,20 @@ export class DatasetClient<
...
+                method: 'GET',
+                params: this._params(datasetListOptions),
+                timeout: DEFAULT_TIMEOUT_MILLIS,
+            });
+
+            return this._createPaginationList(response, datasetListOptions.desc ?? false);
+        };
 
-        return this._createPaginationList(response, options.desc ?? false);
+        return this._listPaginatedFromCallback(listItems.bind(this), options);
```
barjin: JS nit : since `const listItems()` is an arrow function, it holds `this` reference in closure (and bind is noop on it):

image

image

barjin: (tldr, you can drop `.bind()` here :))

PR #783, opened 2025-11-13
docs: Add overview page

marcel-rbro:
Part of the fix for https://github.com/apify/apify-docs/issues/1869

PR #806, opened 2025-12-03
docs: Add versioned overview page

marcel-rbro:
In #783 - forgot to add the overview page to the versioned docs


B4nan:
@marcel-rbro one note for future, please pay attention to the commit titles, for libraries, we do not want docs specific stuff with feat/fix commit types, we dont want those to end up in the generated changelogs. the only exception to this is the apify-docs repo, which doesnt represent a library, but all the rest (clients, sdks, CLI, crawlee, ...) need to follow this rule.

PR #821, opened 2026-01-08
fix: fetch key-value store records as attachments

mvolfik:
See apify/apify-core#25006 - fetching with `attachment=true` now allows getting unmodified HTML, without the script snippet for disabling cookies. There are no other side effects (except that the response will have `Content-Disposition: attachment`, but that is irrelevant to clients other than web browsers)


barjin (approved):
Looks solid to me. Thanks!

PR #835, opened 2026-01-28
feat: Add the new 'ownership' parameter to the storage list methods

nmanerikar:
We introduced a new 'ownership' parameter to the listing endpoints for storages. This PR updates the javascript client to include that parameter.

### DIFF
--- a/src/resource_clients/request_queue_collection.ts
+++ b/src/resource_clients/request_queue_collection.ts
@@ -1,6 +1,6 @@
 import ow from 'ow';
 
-import { STORAGE_OWNERSHIP_FILTER } from '@apify/consts'; 
+import { STORAGE_OWNERSHIP_FILTER } from '@apify/consts';
 
 import type { ApiClientSubResourceOptions } from '../base/api_client';
 import { ResourceCollectionClient } from '../base/resource_collection_client';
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -298,6 +298,16 @@ export interface PaginatedList<Data> extends PaginatedResponse<Data> {
  */
 export type PaginatedIterator<T> = Promise<PaginatedList<T>> & AsyncIterable<T>;
 
+/**
+ * Filter for listing storages by ownership.
+ */
+export enum StorageOwnership {
```
tobice: We usually put these inside `apify/apify-shared` from where they are referenced from both external stuff (like clients) and internal stuff (`apify-core`).

See for example [this enum](https://github.com/apify/apify-shared-js/blob/05f472025832c073c37a24bdf3de460dcd4e9218/packages/consts/src/consts.ts#L654) for general resource access.

Sorry, I could have mentioned this in the original `apify-core` PR 😄 But honestly the usual practice is to start in `apify-core` and once the feature stabilizes, move shared stuff out.

tobice (changes requested):

B4nan (approved):
(nothing from my end)

PR #847, opened 2026-02-12
fix: make all node-specific imports `async`, ignore in browser bundle

barjin:
Makes all the Node-specific imports dynamic and fences these off in the `isNode()` condition.

Some popular bundlers and build systems (`vite`, `webpack`) are trying to load the externalized modules, which leads to unexpected errors when importing `apify-client`.

The resulting bundle (after this PR) imports and builds out-of-the-box with default Webpack settings and in a default Vite project.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -172,7 +172,7 @@ export function isNode(): boolean {
  * Use this for Node.js-only modules that should not be included in browser bundles.
  */
 export async function dynamicNodeImport<T = any>(specifier: string): Promise<T> {
-    return await import(specifier)
+    return await import(specifier);
 }
 
 export function isBuffer(value: unknown): value is Buffer | ArrayBuffer | TypedArray {
--- a/test/bundling/package.json
+++ b/test/bundling/package.json
@@ -1,17 +1,17 @@
 {
-  "name": "apify-client-bundling-test",
-  "version": "1.0.0",
-  "main": "index.js",
-  "scripts": {
-    "bundle:webpack": "npm run clean && webpack",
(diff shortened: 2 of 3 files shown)


src/http_client.ts
```diff
@@ -153,7 +104,76 @@ export class HttpClient {
         responseInterceptors.forEach((i) => this.axios.interceptors.response.use(i as any));
     }
 
+    private async ensureNodeInit(): Promise<void> {
+        if (!isNode()) return;
+
+        if (!this.nodeInitPromise) {
+            this.nodeInitPromise = this.initNode();
+        }
```
vladfrangu:
```suggestion
        this.nodeInitPromise ??= this.initNode();
```

src/http_client.ts
```diff
@@ -153,7 +104,76 @@ export class HttpClient {
...
+
+        // Use ProxyAgent which automatically detects proxy from environment variables
+        // and supports CONNECT tunneling
+        const proxyAgent = new ProxyAgent(agentOptions);
+        this.httpAgent = proxyAgent;
+        this.httpsAgent = proxyAgent;
+
+        // Disable Nagle's algorithm for lower latency
+        // This sends data immediately instead of buffering small packets
+        const setNoDelay = (socket: any) => {
```
vladfrangu: Can't we use `socket: import('node:???').IDoNotRememberTheModuleThatHasSockets` for the type, so we dont have lone any's?

barjin: `import type {Socket} from node:net` is removed from the compiled JS anyway, so there's no need for a dynamic import (nor more Webpack tricks).

PR #854, opened 2026-02-25
docs: fix broken links in client docs

barjin:
[related slack thread](https://apify.slack.com/archives/C0L33UM7Z/p1771960770219289)


barjin:
This has been discussed w/ @TC-MO today, as the Docs team wants to unify the docs checks in `apify-docs` and the subprojects. Some sort of reusable workflow used from all the repos is likely the way.

barjin:
fwd @TC-MO do you recall moving any of these links recently? 

I wouldn't bother w/ the API docs links (it's just two endpoints, likely broken by the switch to a different API docs rendering plugin), but the Academy pages might have been linked from other places too

barjin:
To answer @B4nan 's question, I couldn't find the old links in any of our other docs, nor in The Wayback Machine (the correct / new URL for the Academy page [dates back to 2023](https://web.archive.org/web/20230101000000*/https://docs.apify.com/platform/actors/development/programming-interface/environment-variables)).

This leads me to the conclusion that these were some one-off mistakes, and we probably don't have to bother with redirects here.

PR #858, opened 2026-03-02
chore(ci): check for broken links in docs

barjin:
Split off from #854 

Adds automatic CI link checker to `check.yaml` step. If broken links are detected on a PR, the action comments the PR with a link to the link checker summary.


B4nan (approved):
didnt check the new check much, i guess its a copy paste from somewhere else where it works fine, right?

barjin:
We are using the Lychee checker with a very similar setup in `apify-docs`, so yes, that one is mostly copy-pasted.

If you want to see how it looks when it fails, check [this comment's history](https://github.com/apify/apify-client-js/pull/858#issuecomment-4623140088). 

See that if everything passes in the beginning, there's no comment not to spam the thread too much.

szaganek (approved):
Amazing 😎

PR #876, opened 2026-03-26
fix: Turn any log streaming related errors to warnings

Pijukatel:
### Description
Errors in log streaming should not propagate further. Just log them as warnings.
Such errors should be fixed on API level: https://github.com/apify/apify-core/issues/26653

### Issues
Closes: #873

### DIFF
--- a/src/resource_clients/log.ts
+++ b/src/resource_clients/log.ts
@@ -4,7 +4,7 @@ import type { Readable } from 'node:stream';
 import c from 'ansi-colors';
 
 import type { Log } from '@apify/log';
-import { Logger, LogLevel } from '@apify/log';
+import log, { Logger, LogLevel } from '@apify/log';
 
 import type { ApifyApiError } from '../apify_api_error';
 import type { ApiClientSubResourceOptions } from '../base/api_client';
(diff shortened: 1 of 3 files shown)


src/resource_clients/log.ts
```diff
@@ -197,11 +197,15 @@ export class StreamedLog {
...
-            this.destinationLog.info(lastMessage);
+        try {
+            const lastChunkRemainder = await this.logStreamChunks(logStream);
+            // Process whatever is left when exiting. Maybe it is incomplete, maybe it is last log without EOL.
+            const lastMessage = Buffer.from(lastChunkRemainder).toString().trim();
+            if (lastMessage.length) {
+                this.destinationLog.info(lastMessage);
+            }
+        } catch (err) {
+            this.destinationLog.warning(`Log redirection stopped due to error: ${JSON.stringify(err)}`);
```
barjin: There is a second `data` parameter in all(?) `Log` methods ([code](https://github.com/apify/apify-shared-js/blob/0e14c72b2bbea8117bf543887b42c32984d31a6c/packages/log/src/log.ts#L295)), where you can pass structured data which gets serialized using the logger-wide rules. This might help with the `err` printing (JSON serialization just makes everything an opaque, one-line string)

barjin (approved):
Nice, I didn't even hope it would be this simple 😄 

I have one nit regarding the warning message formatting, otherwise lgtm. Thank you @Pijukatel !

src/resource_clients/log.ts
Pijukatel: Good point. Also, it made me realize that this should be logged by the normal logger and not the redirect one, because this is not the log from the redirected actor run.

PR #879, opened 2026-03-31
fix: pin axios to 1.14.0 — supply chain attack mitigation

paulonasc:
## Summary

- Pins `axios` direct dependency from `^1.6.7` to exact `1.14.0`

## Context

`axios@1.14.1` and `axios@0.30.4` were compromised with a malicious dependency (`plain-crypto-js@4.2.1`) that deploys a Remote Access Trojan (RAT) via a postinstall hook. The RAT targets macOS, Linux, and Windows and phones home to a C2 server.

Anyone running `npm install` with `apify-client` currently resolves `axios@^1.6.7` to the compromised `1.14.1`.

**Advisory:** https://www.aikido.dev/blog/axios-npm-compromised-maintainer-hijacked-rat

## Changes

- `package.json`: pin axios to exact `1.14.0`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
                 "@crawlee/types": "^3.3.0",
                 "ansi-colors": "^4.1.1",
                 "async-retry": "^1.3.3",
-                "axios": "^1.6.7",
+                "axios": "1.14.0",
                 "content-type": "^1.0.5",
                 "ow": "^0.28.2",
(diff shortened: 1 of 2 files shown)


B4nan:
> Anyone running npm install with apify-client currently resolves axios@^1.6.7 to the compromised 1.14.1.

I don't think this is necessary, the vulnerable version is no longer available on NPM.

PR #880, opened 2026-04-09
fix: add support for new list request-queue requests parameters

mvolfik:
add support for endpoint parameters and new pagination method added in apify/apify-core#12115

There is a new parameter `filter` for the _request queue -> list requests_ endpoint. To properly support pagination, we also added proper cursor-based pagination, with opaque cursor provided by the server. So this PR switches the pagination internals to use this new way. All of this is done in backwards-compatible manner, the test updates are only necessary to update the server implementation mocks and cases where the existing tests were inspecting some internals.

### DIFF
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -683,12 +683,14 @@ export class RequestQueueClient extends ResourceClient {
     ): Promise<RequestQueueClientListRequestsResult> & AsyncIterable<RequestQueueClientRequestSchema> {
         ow(
             options,
-            ow.object.exactShape({
-                limit: ow.optional.number.not.negative,
-                exclusiveStartId: ow.optional.string,
-                cursor: ow.optional.string,
-                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
-            }).validate(mutuallyExclusive('exclusiveStartId', 'cursor'))
+            ow.object
+                .exactShape({
+                    limit: ow.optional.number.not.negative,
+                    exclusiveStartId: ow.optional.string,
+                    cursor: ow.optional.string,
+                    filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
+                })
+                .validate(mutuallyExclusive('exclusiveStartId', 'cursor')),
         );
 
         const getPaginatedList = async (
@@ -786,13 +788,15 @@ export class RequestQueueClient extends ResourceClient {
     ): RequestQueueRequestsAsyncIterable<RequestQueueClientListRequestsResult> {
         ow(
             options,
-            ow.object.exactShape({
-                limit: ow.optional.number.not.negative,
-                maxPageLimit: ow.optional.number,
-                exclusiveStartId: ow.optional.string,
-                cursor: ow.optional.string,
-                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
-            }).validate(mutuallyExclusive('exclusiveStartId', 'cursor')),
+            ow.object
+                .exactShape({
+                    limit: ow.optional.number.not.negative,
+                    maxPageLimit: ow.optional.number,
+                    exclusiveStartId: ow.optional.string,
+                    cursor: ow.optional.string,
+                    filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
+                })
+                .validate(mutuallyExclusive('exclusiveStartId', 'cursor')),
         );
 
         const {
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -12,6 +12,7 @@ import type { ApifyRequestConfig } from '../http_client';
 import {
     cast,
     catchNotFoundOrThrow,
+    mutuallyExclusive,
     parseDateFields,
     pluckData,
     RequestQueuePaginationIterator,
@@ -686,12 +687,9 @@ export class RequestQueueClient extends ResourceClient {
                 limit: ow.optional.number.not.negative,
                 exclusiveStartId: ow.optional.string,
                 cursor: ow.optional.string,
-                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])),
-            }),
+                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
+            }).validate(mutuallyExclusive('exclusiveStartId', 'cursor'))
         );
-        if (options.exclusiveStartId && options.cursor) {
-            throw new Error('Cannot specify both exclusiveStartId and cursor for pagination');
-        }
 
         const getPaginatedList = async (
             rqListOptions: RequestQueueClientListRequestsOptions = {},
@@ -793,9 +791,10 @@ export class RequestQueueClient extends ResourceClient {
                 maxPageLimit: ow.optional.number,
                 exclusiveStartId: ow.optional.string,
                 cursor: ow.optional.string,
-                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])),
-            }),
+                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])).minLength(1),
+            }).validate(mutuallyExclusive('exclusiveStartId', 'cursor')),
         );
+
         const {
             limit,
             exclusiveStartId,
--- a/src/resource_clients/request_queue.ts
+++ b/src/resource_clients/request_queue.ts
@@ -685,6 +685,8 @@ export class RequestQueueClient extends ResourceClient {
             ow.object.exactShape({
                 limit: ow.optional.number.not.negative,
                 exclusiveStartId: ow.optional.string,
+                cursor: ow.optional.string,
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -195,7 +195,7 @@ export function getVersionData(): { version: string } {
 /**
  * Helper class to create async iterators from paginated list endpoints with exclusive start key.
  */
-export class PaginationIterator {
+export class RequestQueuePaginationIterator {
```
mvolfik, line 198: I'm renaming the utility class since it's only used there, and it's not very generic anymore - it requires the underlying API to support explicit cursors, which, for now, only request queue does. AFAICT this class isn't exported publicly from the package, so I hope it's fine?

src/resource_clients/request_queue.ts
```diff
@@ -875,15 +906,21 @@ export interface RequestQueueClientListRequestsOptions {
...
 
 /**
  * Result of listing all requests in the queue.
  */
 export interface RequestQueueClientListRequestsResult {
     limit: number;
+    /** @deprecated Use `cursor` for pagination instead. */
     exclusiveStartId?: string;
+    cursor?: string;
+    nextCursor?: string;
```
barjin, line 926: the API response also seems to return the used `filter` value (if applicable)

Image

barjin (approved):
lgtm, there just seems to be a tiny omission in the return type interface. Thank you @mvolfik !

iirc @Pijukatel / @l2ysho implemented the iterators, maybe they'll have more ideas

l2ysho:
I was more in crawlee iterators, but this looks legit

Pijukatel (commented):
Change seems ok, but I wanted to test more the API behavior and I saw this strange behavior. Is it behaving as expected? Same cursor could be valid/invalid based on the filter parameters
(RQid: `ILOH27Wye6sevTXIA`, cursor: `eyJyZXF1ZXN0SWQiOiIxSE5lT2laZUZ1N2dQMWwifQ`)


Image

mvolfik:
@Pijukatel yep, that's intended - for unfiltered requests, the cursor internally serializes just the request ID (so it's equivalent to how `exclusiveStartId` worked, and we can translate from the old system transparently :) ). For filtered, we are fetching from an index where the unhandled requests are stored, and this index is sorted by `orderNo` (= addition date / lock expiration timestamp). So to continue fetching from this index, we need this `orderNo` in the cursor, and thus the cursor has different shape.

Also, aside from the implementation details, also semantically this behavior makes sense to me. If you fetch half of all results, get a cursor, and then continue fetching only `locked` requests with this cursor, what should it do? Do you have some scenario in mind where it would be practically useful?

src/resource_clients/request_queue.ts
mvolfik: ouch, right, good catch!

this just echoes whatever was provided in the request. However, for the request options, the client accepts array, and serializes it into `?filter=pending,locked`. And the response will contain this string `'pending,locked'`. How should I handle this - just type it as `string`? Or should the client parse this field in the response back into an array? (Or should we fix it on the API level, to return the array? Maybe not a bad idea too)

barjin: Is there any other such endpoint? If so, I'd go for consistency. In the other case, typing it as `string` is imo the cleanest / fastest option, although it might break things for users who do something like

```typescript
const { items, ...rest } = await RequestQueue.listRequests(opts);
/// I want to retry this for some reason
const { items } = await RequestQueue.listRequests(rest); // boom, rest is not the right type, `pending|locked[] !== string`
```

I guess every sane person would have used `opts` in the second call anyway, so the impact should be small.

mvolfik: yeah, that's quite an insane code pattern

the only other endpoint I see with a similar pattern is `/v2/actor-runs?status=FAILED,ABORTED`, which does not echo the status filter in the response at all. I'm not sure what's the reasoning behind this "echo request parameters" pattern, maybe it's just the pagination fields, and we could also drop the filter from the response?

barjin: > we could also drop the filter from the response

Oh yeah, if the API team is happy with this, I'm in favour of this decision too :+1:

Pijukatel:
> Do you have some scenario in mind where it would be practically useful?

Not sure, I had in mind a scenario where the user first iterates using an unfiltered cursor, and at some point, he changes the filter with naive expectation "For the remaining items, iterate with this filter". 

Maybe the implementation-specific behavior could be explained in the spec, so that users will not try the scenario I described above?

Pijukatel (approved):
My comment is not blocking this PR. Seems good to me.

src/resource_clients/request_queue.ts
```diff
@@ -695,7 +700,11 @@ export class RequestQueueClient extends ResourceClient {
                 url: this._url('requests'),
                 method: 'GET',
                 timeout: Math.min(MEDIUM_TIMEOUT_MILLIS, this.timeoutMillis ?? Infinity),
-                params: this._params({ ...rqListOptions, clientKey: this.clientKey }),
+                params: this._params({
+                    ...rqListOptions,
+                    filter: rqListOptions.filter ? rqListOptions.filter.join(',') : undefined,
```
B4nan, line 705: Nit: `rqListOptions.filter ? ...` is truthy for an empty array, so `filter: []` serializes to `filter=` (empty string) on the wire instead of being omitted. Server behavior for an empty `filter` param isn't defined here.

Suggest either:
```ts
filter: rqListOptions.filter?.length ? rqListOptions.filter.join(',') : undefined,
```
or adding `ow.array.minLength(1)` to the validator at line 689 / 796 so an empty array is rejected at the boundary.

src/resource_clients/request_queue.ts
```diff
@@ -685,8 +685,13 @@ export class RequestQueueClient extends ResourceClient {
             ow.object.exactShape({
                 limit: ow.optional.number.not.negative,
                 exclusiveStartId: ow.optional.string,
+                cursor: ow.optional.string,
+                filter: ow.optional.array.ofType(ow.string.oneOf(['locked', 'pending'])),
             }),
         );
+        if (options.exclusiveStartId && options.cursor) {
+            throw new Error('Cannot specify both exclusiveStartId and cursor for pagination');
+        }
```
B4nan: Two small consistency nits:

1. The rest of this client uses `ow` for argument validation (which throws `ArgumentError`); this throws a plain `Error`. Consider encoding the mutual exclusion in the `ow` predicate so the error type matches what callers already handle.
2. `paginateRequests` (~L789–798) defines the same shape but does **not** perform this `exclusiveStartId && cursor` check. The mismatch is reachable: a caller passing both would only fail when the iterator's first page eventually hits `listRequests`. Either mirror the check (or extract it) so both entry points reject early and identically.

test/request_queues.test.ts
```diff
@@ -525,20 +525,26 @@ describe('Request Queue methods', () => {
...
+        test.each([
+            [undefined, undefined],
+            [['pending'], 'pending'],
+            [['pending', 'locked'], 'pending,locked'],
+        ] as const)('listRequests() works', async (filter, filterInQuery) => {
             const queueId = 'some-id';
-            const options = { limit: 5, exclusiveStartId: '123' };
+            const options = { limit: 5, exclusiveStartId: '123' } as RequestQueueClientListRequestsOptions;
+            if (filter) options.filter = filter;
+            const queryForValidation = { ...options, filter: filterInQuery };
```
B4nan: Nit: when `filter` is `undefined`, `queryForValidation = { ...options, filter: undefined }` still carries the `filter` key. This works today because Vitest's `toEqual` ignores `undefined` own-properties (and `optsToQuery` in `test/_helper.ts` doesn't strip them either), but any future tightening — e.g., switching the helper to `toStrictEqual` — would silently break this case. Consider only spreading `filter` into `queryForValidation` when it's defined, mirroring how the test already guards `options.filter`.

test/pagination.test.ts
```diff
@@ -396,7 +402,9 @@ describe('RequestQueueClient.listKeys as async iterable', () => {
                         items,
                         count: items.length,
                         limit: limit || maxItemsPerPage,
-                        exclusiveStartId,
+                        exclusiveStartId: request.params.exclusiveStartId,
+                        cursor: request.params.cursor,
+                        nextCursor: `cursor:${upperIndex}`,
```
B4nan: The mock always returns `nextCursor: cursor:${upperIndex}` — including for the final (empty) page. Because the iterator short-circuits via `items.length === 0` first, this never exercises the new `!page.nextCursor` early-exit branch added in this PR (in both `request_queue.ts` and `utils.ts`).

A test case where the mock returns a non-empty page **without** `nextCursor` would directly cover the new branch and pin down the behavior we're depending on.

B4nan (approved):
Few nits my claude found, nothing blocking.

src/resource_clients/request_queue.ts
mvolfik: dropped https://github.com/apify/apify-core/pull/27557

src/resource_clients/request_queue.ts
mvolfik: added the minlength, thanks!

src/resource_clients/request_queue.ts
mvolfik: nice, good job from claude on the second point. fixed, thanks

PR #886, opened 2026-04-20
chore(deps): update lockfiles to resolve security advisories

B4nan:
Ran `npm update --package-lock-only` + `npm audit fix --package-lock-only` in both the root and `website/` to pick up all in-range security patches. Resolves 35 open [Dependabot alerts](https://github.com/apify/apify-client-js/security/dependabot) across `axios`, `follow-redirects`, `basic-ftp`, `lodash`, `vite`, `node-forge`, `picomatch`, `flatted`, `undici`, `brace-expansion`, `path-to-regexp`, and `yaml`.

Remaining alert #189 (`serialize-javascript` in `website/`) was dismissed as **inaccurate**: Docusaurus is configured with `experimental_faster.rspackBundler: true`, which opts out of the webpack path that pulls in that package. Confirmed by the Docusaurus maintainer in facebook/docusaurus#11801.

Also includes two adjacent fixes needed after the bumps:
- `axios@1.15` widened `response.headers['content-type']` to `AxiosHeaderValue | undefined` — cast to `string` at two call sites.
- `@docusaurus/core@3.10` renamed `future.experimental_faster` to `future.faster`.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -28,7 +28,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [18, 20, 22, 24]
+                node-version: [20, 22, 24]
 
         steps:
             - name: Checkout repository
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,22 +57,22 @@
             }
         },
         "node_modules/@apify/consts": {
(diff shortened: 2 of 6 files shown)


barjin (approved):
lgtm, thanks!

PR #887, opened 2026-04-20
ci: rebase before pushing release commit to avoid non-fast-forward

B4nan:
## Summary
- The release and pre-release workflows wait for code checks to pass before committing the changelog/version bump. If another commit lands on master during that wait, the bot's push is rejected as non-fast-forward (seen in [run 24656479517](https://github.com/apify/apify-client-js/actions/runs/24656479517/job/72091410020)).
- Enables `pull: '--rebase --autostash'` on the `EndBug/add-and-commit@v10` step in both `pre_release.yaml` and `release.yaml` so the release bot reconciles with the latest master before pushing.

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -74,6 +74,7 @@ jobs:
                   author_name: Apify Release Bot
                   author_email: noreply@apify.com
                   message: 'chore(release): Update changelog and package version [skip ci]'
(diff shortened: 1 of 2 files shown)


barjin (approved):
I haven't used these action options before, but a quick Google search results look logical 👍

Thanks!

B4nan:
It was caused by the bump in that action about a month ago, they changed the defaults.

PR #889, opened 2026-05-04
chore: migrate from eslint to oxlint

B4nan:
## Summary

Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter), pulling in the shared `@apify/oxlint-config` preset and `oxlint-tsgolint` for type-aware rules. Mirrors the migration already shipped in apify-storage-local-js, apify-shared-js, and apify-core. Prettier remains in place — the oxfmt swap is the next PR (#TODO).

### What changed

- Drop `eslint`, `typescript-eslint`, `globals`, `eslint-config-prettier`, `@apify/eslint-config`, `@stylistic/eslint-plugin-ts`
- Add `oxlint@1.62.0`, `oxlint-tsgolint@0.22.0`, `@apify/oxlint-config@^0.2.5`
- Replace `eslint.config.mjs` with `oxlint.config.ts`; drop unused `tsconfig.eslint.json`
- `lint` / `lint:fix` now run `oxlint --type-aware`
- Bump `@apify/tsconfig` to `^0.1.2` (drops removed `importsNotUsedAsValues` option that oxlint-tsgolint flagged as invalid)
- Override `tsconfig.json` with `module: Node16`, `moduleResolution: Node16`, and explicit `rootDir: src` so `oxlint-tsgolint` can parse it
- Auto-fix surfaced one regex that should use `String#startsWith` and one unnecessary single-element `Promise.all`

### Lint output

\`\`\`
$ pnpm lint
Found 0 warnings and 0 errors.
Finished in 305ms on 82 files with 116 rules using 14 threads.
\`\`\`

## Test plan

- [x] \`pnpm lint\` — clean
- [x] \`pnpm format:check\` — clean (still on prettier)
- [x] \`pnpm test\` — 352/352 passing
- [x] \`pnpm build\` — node + browser bundles built


oxlint.config.ts
```diff
@@ -0,0 +1,37 @@
...
+                'no-useless-constructor': 'off',
+                'typescript/ban-ts-comment': 'off',
+                'typescript/no-empty-function': 'off',
+                'typescript/no-unused-vars': 'off',
+                'jest/no-conditional-expect': 'off',
+                'vitest/no-conditional-expect': 'off',
+                'jest/expect-expect': 'off',
+                'vitest/expect-expect': 'off',
+                'jest/no-standalone-expect': 'off',
+                'vitest/no-standalone-expect': 'off',
```
barjin, line 39: As in one of the previous PRs - what's the deal [with these rules](https://github.com/vitest-dev/eslint-plugin-vitest/tree/main/docs/rules)? They seem quite reasonable to me 🤔

barjin (approved):
Just one (recurring) question, otherwise lgtm 👍

B4nan:
Re. the vitest test-rules disable block — yeah, the rules themselves are reasonable, but enabling them surfaces ~48 violations and the actual fix is to rewrite the existing `try { await ... } catch (err) { expect(err)... }` pattern to `await expect(...).rejects.toX(...)` everywhere. That's invasive enough that I'd rather keep it as a follow-up so this PR stays scoped to "swap the linter, don't refactor the tests".

Added inline comments to the disable block in 20c557c explaining the same.

PR #890, opened 2026-05-04
chore: migrate to oxfmt

B4nan:
## Summary

Replaces Prettier with [oxfmt](https://oxc.rs/docs/guide/usage/formatter), matching the toolchain in apify-storage-local-js, apify-shared-js, apify-core, apify-web, and crawlee v4.

Stacked on top of #889 (eslint→oxlint migration). Will be retargeted to \`master\` once the parent merges.

### Commits

- \`chore: add oxfmt tooling\` — adds \`oxfmt@0.46.0\` devDep, \`.oxfmtrc.json\` (4-space tabs, single quotes, 120 col), swaps \`format\` / \`format:check\` scripts from prettier to oxfmt, drops \`prettier\`, \`.prettierrc\`, \`.prettierignore\`
- \`ci: rename Prettier step to Format check\` — cosmetic CI tweak

The codebase already matches oxfmt formatting (no reformatting commit needed).

## Test plan

- [x] \`pnpm lint\` — clean
- [x] \`pnpm format:check\` — clean
- [x] \`pnpm build\` — node + browser bundles built


.oxfmtrc.json
```diff
@@ -0,0 +1,15 @@
...
+        "**/*.md",
+        "**/*.json",
+        "**/*.jsonc",
+        "**/*.yaml",
+        "**/*.yml",
+        "**/node_modules",
+        "**/dist",
+        "coverage",
+        "website",
+        "docs"
```
barjin, line 13: There are no TS examples in the client docs, but if there were, wouldn't we want them formatted as well?

This is more about SDK, I suppose (we already merged the same [`oxfmtrc.json` there](https://github.com/apify/apify-sdk-js/blob/master/.oxfmtrc.json), but the docs have some TS files).

barjin (approved):
lgtm here 👍  just a small esprit de l'escalier regarding the SDK changes

.oxfmtrc.json
B4nan: This is aligned with the apify-docs where I disabled formatting of exactly because of examples - many were collapsed into a single line because they would fit the print width, making them effectively less readable. Formatting the markdown should be a follow-up, where we also discuss this with the docs folks (unlike the PRs that are adding oxfmt to the code only, which is not their concern).

PR #895, opened 2026-05-11
fix: drop only-allow preinstall, enforce pnpm via devEngines

### LINKED ISSUE #893: preinstall script in v2.23.2 breaks npm/yarn installs when used as a dependency
apify-client@2.23.2 added "preinstall": "npx only-allow pnpm" to package.json (via #883). This appears to be intended to enforce pnpm for local development, but the script ships in the published npm tarball and fires when apify-client is installed as a dependency of other packages.

  This breaks `npm install -g apify-cli` and any project that depends on apify-client via npm or yarn:
``` bash
  npm error path .../node_modules/apify-client
  npm error command failed
  npm error command sh -c npx only-allow pnpm
```

**Potential fix**:  Either remove the `preinstall` script from the published tarball , or rely on the [`packageManager`](https://nodejs.org/api/packages.html#packagemanager) field + [Corepack](https://github.com/nodejs/corepack) which only applies at the repo level. Note that `only-allow` is [archived](https://github.com/pnpm/only-allow) and has a [known issue](https://github.com/pnpm/only-allow/issues/2) with firing during dependency installation.

  **Affected**: v2.23.2 only (v2.23.1 is fine)

B4nan:
Fixes #893.

## What's wrong

`"preinstall": "npx only-allow pnpm"` ships in the published npm tarball and fires when downstream consumers install `apify-client` via npm/yarn. In the `npm install -g apify-cli` flow the npx bootstrap of `only-allow` fails to put the binary on PATH in time:
- Linux: `sh: 1: only-allow: not found` (exit 127)
- Windows: `'only-allow' is not recognized as an internal or external command` (exit 1)

`only-allow` itself has a `node_modules` guard since v1.x and would exit silently if it ever managed to run, but the `npx` bootstrap fails first in global-install context. The script breaks downstream users for no benefit — local enforcement is already covered by `packageManager` + Corepack + docs.

## What changes

- Drop `preinstall: npx only-allow pnpm`.
- Add `devEngines.packageManager: { name: pnpm, version: 10.24.0, onFail: error }` for hard local-repo enforcement (Node 22+/npm 10+ honors it; older runtimes ignore it harmlessly).
- `devEngines` only fires at the package's own repo root, never on transitive installs, so downstream consumers are unaffected.

Patch release (2.23.3) needed after merge.

## Verified locally

- `npm install apify-client@2.23.2` reproduces the failure in `npm install -g apify-cli` context.
- With this change applied, `pnpm install --frozen-lockfile` still works (no warning, since `packageManager` and `devEngines.packageManager` versions match exactly at `10.24.0`).
- `npm install` in this repo now errors with `EBADDEVENGINES` instead of pretending to work — same outcome as before, cleaner signal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,7 @@
         "format": "oxfmt",
         "format:check": "oxfmt --check",
         "build:node": "tsc",
-        "build:browser": "rsbuild build",
-        "preinstall": "npx only-allow pnpm"
+        "build:browser": "rsbuild build"
     },
     "dependencies": {
         "@apify/consts": "^2.50.0",
@@ -108,5 +107,12 @@
         "webpack": "^5.105.2",
         "webpack-cli": "^7.0.0"
     },
-    "packageManager": "pnpm@10.24.0"
+    "packageManager": "pnpm@10.24.0",
+    "devEngines": {
+        "packageManager": {
+            "name": "pnpm",
+            "version": "10.24.0",
+            "onFail": "error"
+        }
+    }
 }


vladfrangu:
Will it work now? I recall you saying it wouldn't because pnpm delegates to npm internally which is why we couldn't use this

B4nan:
Let's see, the problem we were facing before was `pnpm config` calling `npm config`, but we don't use that here. If this won't work, we should see it in the canary release. I was trying a dry run locally and `pnpm publish` doesn't call `npm` directly (it uses `@npmcli/*` dependency for that).

vladfrangu (approved):
🤞

PR #897, opened 2026-05-11
chore: bump pnpm to 10.33.4

B4nan:
Brings `packageManager` and `devEngines.packageManager.version` up to the latest pnpm v10 patch (was 10.24.0). No behavior change — just keeps the version pin current.

Same bump going across the other recently-migrated public repos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,11 +107,11 @@
         "webpack": "^5.105.2",
         "webpack-cli": "^7.0.0"
     },
-    "packageManager": "pnpm@10.24.0",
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):
fwiw I think we have version ranges in devEngines but this is fine too (plus we're bumping to 11 soon right?)

B4nan:
> plus we're bumping to 11 soon right?

Not really soon, but hopefully by the end of the year, we'll wait for major bump with this since v11 requires node 22.

PR #901, opened 2026-05-15
docs: use the common `ApiLink` component

barjin:
Uses the common `ApiLink` component from the shared `@apify/docs-theme`. 

Rewrites the existing documentation to use this component.


B4nan (approved):
nit: do we really need to rename the files to mdx?

barjin:
> do we really need to rename the files to mdx

Apparently not, Docusaurus seems to use the MDX parser anyway (and VSCode's MD syntax parser seems to understand the `<>` tags already). Thanks for pointing this out :+1:

PR #925, opened 2026-05-29
chore: stop publishing broken declaration and source maps

B4nan:
The published tarball excludes `src` but ships `.d.ts.map`/`.js.map` that reference `../src/*.ts`, so the maps never resolve — editors fall back to the compiled `.d.ts`/`.js` instead of the original source. Disabling `declarationMap`/`sourceMap` stops emitting these dead maps and shrinks the tarball. The rsbuild browser bundle sourcemap (which embeds its own sources) is unaffected.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,9 @@
         "moduleResolution": "Node16",
         "outDir": "dist",
         "rootDir": "src",
-        "skipLibCheck": true
+        "skipLibCheck": true,
+        "declarationMap": false,
+        "sourceMap": false
     },
     "include": ["./src/**/*"]
 }


barjin (approved):
I see the branch doesn't contain the changes from https://github.com/apify/apify-client-js/pull/919, so the failing checks are probably expected.

The changes in this branch look alright to me. Thanks @B4nan !

B4nan:
Yeah, the CI is unrelated, I guess me enabling renovate on 10 of our repos made quite a huge hit to both pw and pup servers :D I hope things will stabilize once I merge/close all those PRs, also added cache for pw browsers (but we need to update to 1.60 anyway).

PR #936, opened 2026-06-11
docs: add `@since` TypeDoc tags, drop versioned documentation

barjin:
Drops the Docusaurus-based versioned documentation and adds the `@since` tags to the JSDoc blocks. These are rendered as shown below :arrow_down: 

image

The main issue remains the need for manual updates of the `@since` tags on new symbols. The initial seeding was done by a script - perhaps we could add this to the CI?

----

This is an experimental PR as a follow-up to an internal Tooling team discussion.


src/resource_clients/actor_collection.ts
```diff
@@ -89,41 +92,125 @@ export class ActorCollectionClient extends ResourceCollectionClient {
...
+ * @since Added in 2.0.1
+ */
 export interface ActorCollectionListItem {
+    /**
+     * @since Added in 2.0.1
+     */
     id: string;
+    /**
+     * @since Added in 2.0.1
+     */
```
janbuchar: This seems redundant — no way the property could've been added before the interface itself... right?

barjin: Yes... but I'd have everything explicit rather than implementing some unclear "distributive" logic in the plugin. The `@since` tag in TypeDoc doesn't have any deeper logic behind it ([docs](https://typedoc.org/documents/Tags._since.html)).

Some of the docblocks now look rather repetitive; this is something I'm hoping will change with the OpenAPI-generated clients (and the docs team's initiative to have better docblocks, as briefly discussed recently).

janbuchar: Can we at least hide these repeated ones in the rendered docs?

barjin:
This is what the page from above looks like after recent changes:

image

`@since` tags are basically only on the necessary spots (top-level symbols and later additions) - both in code and rendered.

Wdyt @janbuchar @B4nan ? 


(PS: note that small mismatches can happen because of the automatic annotation script. The `@since` tags in code can be freely updated by hand / agent)

src/resource_clients/actor_collection.ts
barjin: See https://github.com/apify/apify-client-js/pull/936#issuecomment-4741795431

janbuchar (approved):
I like it, thanks!

src/resource_clients/actor.ts
```diff
@@ -456,64 +470,141 @@ export class ActorClient extends ResourceClient {
  * Actors are serverless computing units that can perform arbitrary tasks such as web scraping,
  * data processing, automation, and more. Each Actor has versions, builds, and can be executed
  * with different configurations.
+ * @since Added in 0.5.0
```
B4nan: i thought we dont really need this if it was part of the current major version from the beginning?

src/resource_clients/actor.ts
```diff
@@ -456,64 +470,141 @@ export class ActorClient extends ResourceClient {
...
  */
 export interface Actor {
-    /** Unique Actor ID */
+    /**
+     * Unique Actor ID
+     * @since Added in 2.0.1
```
B4nan: there is no v2.0.0 tag, so 2.0.1 is the first version. i was hoping we wont have those tags on all properties, and i guess this is the main reason why they're there.

website/docusaurus.config.js
```diff
@@ -63,17 +62,6 @@ module.exports = {
                             href: 'https://github.com/apify/apify-client-js',
                             label: 'GitHub',
                         },
-                        {
-                            type: 'docsVersionDropdown',
```
B4nan, line 67: i thought we agreed to keep majors versioned as before, so we still want the version switcher

B4nan (commented):
we should add the claude.md here and mention this there, otherwise i wouldnt trust the agents to do this consistently. also its a bit unfortunate that now you need to write a version before it gets released (usually it will be a feature bump, given new options/methods are added, but not always).

B4nan:
btw can we also render the `@deprecated` tags same way? even better, lets render all the tags the same?

src/resource_clients/actor.ts
barjin: good point, solved in https://github.com/apify/apify-client-js/pull/936/commits/899121f0914121def5035c99e4d23011c0821982

src/resource_clients/actor.ts
barjin: also solved in https://github.com/apify/apify-client-js/pull/936/commits/899121f0914121def5035c99e4d23011c0821982 . The `@since` tags are now only on post-`2.0.1` additions.

website/docusaurus.config.js
barjin: Now, this is a fair discussion point.

The _current_ client docs only have the `latest` and `next` versions. Since the last v1 version is 5 years old (and has ~100 downloads, compared to 200k downloads on the v2 latest), I don't think it makes much sense to revive the v0 and v1 docs for this project. Perhaps for other projects, it would make more sense.

> otherwise i wouldnt trust the agents to do this consistently

I have made a script that adds the `@since` tags from TypeDoc and `git` tags, so it's fairly deterministic.

> its a bit unfortunate that now you need to write a version before it gets released

See the original PR description - perhaps we could run the aforementioned script in CI and have it create a PR with the `@since` tags on release?

---

Regarding versioning, should we keep the `next` version to keep the new symbols away from the main live docs? Or should we disable versioning altogether, write `@since Added in next` manually to each new symbol, and have it replaced on release by the script from above?

Also (`apify-client-js` specific) - could the autogenerated clients solve any of this for us in the future?

barjin:
> btw can we also render the `@deprecated` tags same way? even better, lets render all the tags the same?

Yes, here's [the issue](https://github.com/apify/docusaurus-plugin-typedoc-api/issues/85).

barjin:
The recent changes to the Typedoc plugin update rendering of the other tags as well (see `@deprecated` section below)

image

barjin:
@janbuchar do you think we could use the API extractor (that's already in, e.g., Crawlee) to find new symbols in a new release and auto-annotate these with the `@since` tags? Does that sound feasible?

PR #948, opened 2026-06-23
chore: bump `docusaurus-plugin-typedoc-api` version

barjin:
Improves rendering of @-prefix tags in the documentation.

See discussion under https://github.com/apify/apify-client-js/pull/936 for more info.


barjin:
For the reviewers - this is here to get the quick rendering wins before we decide on the `@since` tags from #936

B4nan (approved):

PR #952, opened 2026-06-26
fix: Fix incomplete type hint of last run/task

Pijukatel:
The API accepts the `origin` argument as well. Type is not complete.
- Fix type by adding `origin`
- Reuse duplicate type
- Add the argument to one integration test to make it covered.

API source:
Task:
https://github.com/apify/apify-core/blob/v0.1508.0/src/api/src/routes/actor_tasks/last_run.ts#L21
Run:
https://github.com/apify/apify-core/blob/v0.1508.0/src/api/src/routes/actors/last_run.ts#L21

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -845,7 +845,7 @@ export interface ActorBuildOptions {
  */
 export interface ActorLastRunOptions {
     status?: ValueOf<typeof ACTOR_JOB_STATUSES>;
-    origin?: ValueOf<typeof ACTOR_JOB_STATUSES>;
+    origin?: ValueOf<typeof META_ORIGINS>;
 }
 
 /**
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -2,7 +2,7 @@ import type { AxiosRequestConfig } from 'axios';
 import ow from 'ow';
 
 import type { RUN_GENERAL_ACCESS } from '@apify/consts';
-import { ACT_JOB_STATUSES, ACTOR_PERMISSION_LEVEL, META_ORIGINS } from '@apify/consts';
+import { ACTOR_JOB_STATUSES, ACTOR_PERMISSION_LEVEL, META_ORIGINS } from '@apify/consts';
 import { Log } from '@apify/log';
 
 import type { ApiClientSubResourceOptions } from '../base/api_client';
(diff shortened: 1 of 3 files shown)


Pijukatel:
Found when inspecting the generated client update that added an undocumented endpoint just because of the JS implementation:
https://github.com/apify/apify-client-rust/pull/8

src/resource_clients/actor.ts
```diff
@@ -844,6 +844,7 @@ export interface ActorBuildOptions {
  */
 export interface ActorLastRunOptions {
     status?: keyof typeof ACT_JOB_STATUSES;
```
barjin: See that the type is wrong in `status` as well.

The type asks for `keyof typeof`, which are the object _keys_, but the later `ow` validation asks for `Object.values(ACT_JOB_STATUSES)`) (the object _values_).

https://github.com/apify/apify-client-js/blob/59afcf09c6f01027033fccdaf72d4e92c2049c20/src/resource_clients/actor.ts#L375

This is fine for most statuses, but not `TIMING_OUT: 'TIMING-OUT'` and `TIMED_OUT: "TIMED-OUT"` ([see source](https://github.com/apify/apify-shared-js/blob/d31388fa0a48b8a06263ba973f5bd584ab0c26b3/packages/consts/src/consts.ts#L35))

This means we can get either a TS compilation error (but no runtime errors) like this:

Image

Or no TS compilation error, but a validation runtime error like this:

Image

Image

barjin: Fortunately, the `keys === values` in `META_ORIGINS` :)

barjin: See that `ACT_JOB_STATUSES ` is [deprecated](https://github.com/apify/apify-shared-js/blob/d31388fa0a48b8a06263ba973f5bd584ab0c26b3/packages/consts/src/consts.ts#L186) and we should use ACTOR_JOB_STATUSES instead

barjin (approved):
Thank you @Pijukatel !

The changes are imo right, so approving 👍  Reading this, I noticed another small mismatch - feel free to merge now and create this into an issue (although the solution should be a few-liner).

src/resource_clients/actor.ts
Pijukatel: Using ACTOR_JOB_STATUSES

Pijukatel: Using values now

PR #964, opened 2026-07-09
fix: align User-Agent OS token with other Apify clients

Pijukatel:
## What & why

The JS client builds its `User-Agent` OS token from `os.type()`, which returns `uname`-style names (`Linux`, `Darwin`, `Windows_NT`). The Python client uses `sys.platform`, which returns short lowercase identifiers (`linux`, `darwin`, `win32`). This makes the same platform look different across clients — most starkly on Windows (`Windows_NT` vs `win32`).

This PR switches JS from `os.type()` to `os.platform()`. Node's `os.platform()` returns exactly the same lowercase identifiers as Python's `sys.platform`, so the OS token lines up across clients on every OS with no per-OS special-casing.

## OS token before/after

| OS | `os.type()` (before) | `os.platform()` (after) | Python `sys.platform` |
|----|----------------------|-------------------------|-----------------------|
| Linux | `Linux` | `linux` | `linux` |
| Android | `Linux` | `android` | `android` |
| macOS | `Darwin` | `darwin` | `darwin` |
| Windows | `Windows_NT` | `win32` | `win32` |

## Changes

- `src/http_client.ts`: `os.type()` → `os.platform()` in the User-Agent string.
- `test/http_client.test.ts`: assert the OS token matches `os.platform()`.

## Notes

Example of current data that is inconsistent

image

### DIFF
--- a/src/http_client.ts
+++ b/src/http_client.ts
@@ -159,7 +159,7 @@ export class HttpClient {
 
         // Works only in Node. Cannot be set in browser
         const isAtHome = !!process.env[APIFY_ENV_VARS.IS_AT_HOME];
-        let userAgent = `ApifyClient/${version} (${os.type()}; Node/${process.version}); isAtHome/${isAtHome}`;
+        let userAgent = `ApifyClient/${version} (${os.platform()}; Node/${process.version}); isAtHome/${isAtHome}`;
 
         if (this.userAgentSuffix) {
             userAgent += `; ${asArray(this.userAgentSuffix).join('; ')}`;
--- a/test/http_client.test.ts
+++ b/test/http_client.test.ts
@@ -1,4 +1,5 @@
 import type { AddressInfo } from 'node:net';
+import os from 'node:os';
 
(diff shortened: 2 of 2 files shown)


Pijukatel:
Effort to make the user agents' headers consistent for our clients.
Previous alignemnt in Python client: https://github.com/apify/apify-client-python/pull/903

barjin (approved):
lgtm, thank you @Pijukatel !

fwiw, `process.version` will print, e.g., `v24.10.1` (with the `v` prefix). Python [doesn't do this](https://docs.python.org/3/library/sys.html#sys.version_info)

Pijukatel:
> lgtm, thank you @Pijukatel !
> 
> fwiw, `process.version` will print, e.g., `v24.10.1` (with the `v` prefix). Python [doesn't do this](https://docs.python.org/3/library/sys.html#sys.version_info)

That is ok from data processing point of view as  Python and Node will never share version literals that need to be grouped in the same processing. (As opposed to OS or isAtHome)

PR #967, opened 2026-07-13
feat: support APIFY_API_BASE_URL and APIFY_API_PUBLIC_BASE_URL env vars

Pijukatel:
## Description

The API base URL and public API base URL were hardcoded as inline literals in the `ApifyClient` constructor. Anyone pointing the client at a local platform, a staging deployment, or a custom host/port had to edit or wrap the library.

Both URLs now resolve as **explicit constructor option > environment variable > built-in default constant**:

- `APIFY_API_BASE_URL` → `baseUrl`
- `APIFY_API_PUBLIC_BASE_URL` → `publicBaseUrl`

The two variables are independent (each keeps its own default). Empty-string env values are treated as unset (fall through to the default). The production defaults are now top-level constants, env reads are browser-safe (guarded so they don't throw when `process` is undefined), and the automatic `/v2` suffixing is unchanged. **Setting nothing reproduces today's behavior exactly.**

### Context 

This is one of three sibling PRs applying the same env-var contract across the Apify clients and CLI (the CLI also adds `APIFY_CONSOLE_BASE_URL`). Separate repos, no overlapping files:

- apify-client-js (this PR): apify/apify-client-js#967
- apify-client-python: apify/apify-client-python#948
- apify-cli: apify/apify-cli#1275

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,23 @@ Some actions can't be performed by the API itself, such as indefinite waiting fo
 Key-value store records can be retrieved as objects, buffers or streams via the respective options, dataset items
 can be fetched as individual objects or serialized data and we plan to add better stream support and async iterators.
 
+### Custom API URLs
+
+By default, the client talks to `https://api.apify.com`. You can point it elsewhere — e.g. a local or staging
+instance of the Apify platform — either via constructor options or environment variables:
+
+```js
+const apifyClient = new ApifyClient({
+    token: 'my-token',
+    baseUrl: 'http://localhost:8080',
+});
+```
+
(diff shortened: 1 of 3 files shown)


barjin (commented):
See that this is handled in SDK already ([source code](https://github.com/apify/apify-sdk-js/blob/02057036f324ef2e3f6b952d2b734918943db7f1/src/configuration.ts#L207)). 

iirc, this is because we don't want the Apify Client to care about its run environment (it just runs anywhere, even in browsers). Reading (all) the envvars was the SDK's job, as that's made specifically for writing Actors (that will run on Apify).

Pijukatel:
Close in favor of smaller CLI update: https://github.com/apify/apify-cli/pull/1276

PR #983, opened 2026-07-30
chore: vendor the OpenAPI spec and add a refresh script

vdusek:
Vendors the published OpenAPI spec as a committed snapshot, plus a script to refresh it. Input artifact only — nothing reads it yet. Precondition for #985, which generates all output models from this snapshot and refreshes it on a schedule; that PR lands on top of this one.

- **`spec/openapi.json`** — the snapshot (995 kB, `openapi: 3.1.2`, 131 paths, 237 schemas), and the only input codegen will read. `linguist-generated`, so it collapses in diffs.
- **`scripts/fetch-spec.mts`** + `pnpm spec:fetch` — downloads it, rejects anything that isn't a plausible OpenAPI 3.1 document, then writes the bytes verbatim via temp file + rename. An error page served with a 200 can't overwrite the snapshot.
- **`tsconfig.scripts.json`** + `pnpm tsc-check-scripts` — `scripts/` sat outside the main project's `include`, so nothing type-checked it. Wired into the existing `lint` job.
- **`.gitattributes`** — pins LF repo-wide, so a clone made with `core.autocrlf=true` can't turn a refresh into 29k lines of line-ending churn. Renormalizes nothing today.
- **`spec/` and not `src/`** — it's a build input, not source. `resolveJsonModule` is on, so one stray import from `src/` would emit the whole document into `dist/` and publish it (verified). Same layout as the Python client; the generated types will land in `src/generated/api.ts`.
- **Don't review the JSON diff** — run `pnpm spec:fetch` and confirm the tree stays clean. That proves the committed bytes are what the endpoint serves.
- **`info.version` is an apify-docs build stamp**, not an API version, so any docs redeploy flips that one line. The sync cron will have to gate on the generated output rather than on a spec diff.
- **`pnpm spec:fetch` needs Node >= 22.18** (native type stripping). Maintainer-only, so no `engines` change — that belongs with the v4 Node baseline.

*✍️ Drafted by Claude Code*


vdusek:
Superseded by #985, which folds this in. The fetch script moved there; the vendored snapshot is gone -- the specification is now downloaded into git-ignored `tmp/` at generation time and only its version is committed, matching apify-client-python.

*✍️ Drafted by Claude Code*

PR #984, opened 2026-07-30
chore!: require Node.js 22 or newer

vdusek:
Raises the minimum supported Node.js version to 22, matching the Crawlee v4 floor. `apify-client` sits below `@crawlee/core`, so its minimum must not be higher than Crawlee's, and there is no reason for it to be lower.

- **`package.json`** — added `engines: { "node": ">=22.0.0" }`.
- **`check.yaml`** — `build_and_test` matrix `[20, 22, 24]` -> `[22, 24, 26]`, so the client is now also tested against Node.js 26. The remaining jobs stay on 24, as Node.js 26 does not become LTS until October 2026.
- **`tsconfig.json`** — dropped the `module: "Node16"` and `moduleResolution` overrides, so the `@apify/tsconfig` base values (`node20`/`nodenext`) apply; `lib` -> `["ES2024", "DOM"]`, `target: "ES2024"`. ES2024 rather than Crawlee's `ESNext`, because Node 22 does not implement ES2025 syntax such as `using`.
- **Docs** — updated `docs/01_introduction/index.md` and `CONTRIBUTING.md`. `website/versioned_docs/version-2/**` is untouched, as it is a frozen v2 snapshot.

Drive-by: `pnpm tsc-check-tests` was already failing on `v3`, mostly with `TS1541` from the old `module: Node16`. Nothing ran it in CI, so it went unnoticed; it is fixed and now part of the `lint` job.

BREAKING CHANGE: Node.js 18 and 20 are no longer supported. The minimum supported version is now Node.js 22.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -28,7 +28,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [20, 22, 24]
+                node-version: [22, 24]
 
         steps:
             - name: Checkout repository
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -28,7 +28,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [22, 24]
+                node-version: [22, 24, 26]
 
         steps:
             - name: Checkout repository
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -79,6 +79,8 @@ jobs:
             - name: Install pnpm and dependencies
               uses: apify/actions/pnpm-install@v1.4.0
             - run: pnpm lint
+            - name: Type-check tests
+              run: pnpm tsc-check-tests
             - name: Format check
               run: pnpm format:check
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -119,8 +119,8 @@ async function brotliValue(value: string | Buffer<ArrayBufferLike>): Promise<Buf
         const { promisify } = await import('node:util');
         const { brotliCompress, constants } = await import('node:zlib');
         const compress = promisify(brotliCompress);
-        brotliCompressPromisified = async (value) =>
-            compress(value, { params: { [constants.BROTLI_PARAM_QUALITY]: 6 } });
+        const options = { params: { [constants.BROTLI_PARAM_QUALITY]: 6 } };
+        brotliCompressPromisified = async (input) => compress(input, options);
     }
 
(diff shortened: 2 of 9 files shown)


.github/workflows/check.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                node-version: [20, 22, 24]
+                node-version: [22, 24]
```
B4nan:
```suggestion
                node-version: [22, 24, 26]
```

src/utils.ts
```diff
@@ -109,46 +109,36 @@ export function stringifyWebhooksToBase64(webhooks: WebhookUpdateData[]): string
     return btoa(String.fromCharCode(...uint8Array));
 }
 
-let gzipPromisified: ((arg: string | Buffer<ArrayBufferLike>) => Promise<Buffer>) | undefined;
+let brotliCompressPromisified: ((arg: string | Buffer<ArrayBufferLike>) => Promise<Buffer>) | undefined;
```
B4nan: isnt this all unrelated to the nodejs bump?

.github/workflows/check.yaml
vdusek: I thought we were gonna add support for the new Node version in a follow-up PR, because of the changelog, but I guess it's fine with the title "Requiring Node.js 22 or newer".

src/utils.ts
vdusek: Well, there was the old guard for brotli compression with a `typeof brotliCompress === 'function'` check, whose comments said it was a workaround for Node < 10.16.0 :smile:. That should be dead code for Node 22+, right? It probably slipped through the review when adding brotli support.

And I transformed the encoding into a real try/catch block with gzip fallback if brotli compression fails, for runtimes that do not ship brotli.

Let me know WDYT. I can either revert all the changes here and keep the version guard there, or keep it as it is. No problem with either option.

.github/workflows/check.yaml
B4nan: i dont think we need to do anything to "add support for node 26", we just need to test it, so it doesnt really need a changelog entry

src/utils.ts
B4nan: on the other hand, this one sounds like it deserves its own PR

vdusek: removed, dedicated PR for it -> https://github.com/apify/apify-client-js/pull/990

B4nan (approved):
simplifed the tsconfig since it was just mirroring what the base one already set up

PR #985, opened 2026-07-30
feat!: generate all output models from the OpenAPI spec

vdusek:
Closes:  #1026 

## Description

- API response types are now generated from the OpenAPI specification instead of hand-written, using [openapi-typescript](https://openapi-ts.dev/).
- `pnpm generate:types` downloads the specification into git-ignored `tmp/` and writes `src/generated/api.ts`. Only the specification version is committed, in `package.json`.
- Nothing re-exports the generated file: `src/models.ts` declares each published model on top of a generated schema, and `src/spec_guards.ts` asserts every deviation at compile time.
- A nightly CI workflow regenerates and opens a pull request if anything changed.
- It follows the same approach as the Python API client.

## Breaking

- Several types were outright wrong, and many fields gained null or became optional to match what the API actually returns.
- Two runtime changes: `parseDateFields()` depth 3 -> 4, and the key-value store's next-key check widens to `!= null`.
- Everything is described in the v3 upgrading guide.

## Downstream

- apify/apify-sdk-js#702 runs the v4 SDK against this branch: six type errors, all absorbed by v4's backend adapters. Build, typecheck, unit tests, lint and format pass.
- `handledAt` and `retryCount` break only on the v3 line, where the SDK relies on structural `StorageClient` compatibility.

## Open

- `notify_on_failure` needs a `SLACK_WEBHOOK_URL` secret this repo does not have.

*✍️ Drafted by Claude Code*


vdusek:
Regarding the single generated file size; `openapi-typescript` itself doesn't support splitting one spec's generated types into multiple files.

What we can do:
- add the generated file to the ignores in `oxlint.config.ts`, to not affect linter performance;
- add it to the `.gitattributes`, to always hide the diff in GitHub;
- regarding the language server, there shouldn't be much impact, according to measurements by Claude:

> Full clean tsc --noEmit over the whole project (485 files) takes 0.89s. Per-file trace attribution shows generated/api.ts is the single most expensive file in the program at 78ms (11ms parse + 20ms bind + 46ms check) — about 10% of the summed per-file cost, but under 0.1s in absolute terms. For comparison, tiny hand-written files like http_client.ts (53ms check) and interceptors.ts (49ms check) have comparable or higher check cost from generic/retry-decorator logic despite being a few hundred lines — proving check cost tracks type complexity, not line count.

Let me know @B4nan WDYT.

B4nan:
Let's leave it, it's true that it's just a huge file, but the code is far away from complex.

vdusek:
Let's wait for https://github.com/apify/apify-client-js/pull/995 and then rebase upon it, so that we can better test this.

package.json
```diff
@@ -49,10 +49,12 @@
     "homepage": "https://docs.apify.com/api/client/js/",
     "files": [
         "dist",
-        "!dist/*.tsbuildinfo"
+        "!dist/*.tsbuildinfo",
+        "!dist/spec_guards.*"
     ],
     "scripts": {
         "build": "pnpm clean && pnpm build:node && pnpm build:browser",
+        "prepare": "npm run clean && npm run build:node && npm run build:browser && npm run postbuild",
```
B4nan: `apify-client@3.0.0-beta.0` is out so we can drop this

barjin (approved):
I don't see any major issues, so approving. Thanks @vdusek !

One point about the generated documentation - the OpenAPI tags will leak into the generated API docs, we should render them better. I made an issue about this [here](https://github.com/apify/docusaurus-plugin-typedoc-api/issues/90)

B4nan (approved):
looks good to me as well

PR #986, opened 2026-07-30
refactor!: validate arguments with zod instead of ow

vdusek:
Closes:  #1024 

Replaces `ow` with `zod` for runtime argument validation. Input validation only — response validation is a separate PR.

## How it works

- `ArgumentValidationError` and `validate()` live in this package: `apify-client` sits below `@crawlee/core` and the SDK in the dependency graph, so it cannot import theirs.
- Schemas use `z.strictObject` / `z.looseObject` / `z.enum`, not the deprecated `.strict()` / `.passthrough()` / `z.nativeEnum()`.
- `chunkSize` now works on every paginating `list()` whose schema spreads the shared `paginationOptionsShape`. It used to type-check but throw; `ow`'s `exactShape` had the same gap.

## Browser bundle

- Tree-shaking and minification are back on in `rsbuild.config.ts`, off since the webpack-to-rsbuild migration in #671. Now 288 kB raw / 87 kB gzip, from 1439 kB / 273 kB — below the 946 kB / 203 kB before this PR. A 320 kB budget fails the build, so it cannot grow unnoticed again.

## Breaking changes

- Invalid arguments throw `ArgumentValidationError` (exported from `apify-client`), not `ow`'s `ArgumentError` — different messages, the zod issues on `issues`, the original `ZodError` on `cause`. A message renders at most 10 problems, then `... and N more`.
- Arrays and functions no longer pass where a plain object is expected: `update()` / `create()` fields, `TaskClient.start()` / `call()` input, the storage `schema` option, `DatasetClient.pushItems()` items, `RequestQueueClient.addRequest()` / `batchAddRequests()` requests.
- `Infinity` no longer passes on numeric options such as `waitSecs`, `timeout` or `memory`, and an invalid `Date` no longer passes on `startedBefore` / `startedAfter` — `z.number()` requires a finite number and `z.date()` a valid date, where `ow` only checked the type.
- `KeyValueStoreClient.setRecord()` rejects `Infinity` as a record value, which v2 accepted and stored as `null`.
- Options that were declared but always rejected at runtime are gone from the types: `chunkSize` on `downloadItems()` and `createItemsPublicUrl()`, `signature` on `createItemsPublicUrl()` and `createKeysPublicUrl()` — a compile error now instead of a throw.
- `Date`, `Map`, `Set` and other class instances still pass as objects, as under `ow`.

*✍️ Drafted by Claude Code*


szaganek (approved):

B4nan:
> Follow-ups: the browser bundle grows from 946 kB to 1446 kB raw (203 kB -> 272 kB gzip), because rsbuild.config.ts disables tree-shaking and minification; .strict(), .passthrough() and z.nativeEnum() are deprecated in zod 4 and could move to z.strictObject / z.looseObject / z.enum; chunkSize is missing from every .strict() list schema (pre-existing — ow's exactShape had the same gap).

I would rather fix it here before it gets merged. We don't want to use any deprecated methods, and this PR introduces the bundle size issue.

vdusek:
> I would rather fix it here before it gets merged. We don't want to use any deprecated methods, and this PR introduces the bundle size issue.

OK, I'll check it out

vdusek:
@B4nan it's ready for a re-check

vdusek:
I adopted the approach & naming from https://github.com/apify/crawlee/pull/3935 (~`validate`~, `parseArgument`). @vladfrangu could you recheck, please?

vdusek:
I'm merging it into the `v3` branch. If you guys find anything, I can address it later.

PR #989, opened 2026-08-03
feat: add Actor task publication endpoints

Janjiran:
Part of https://github.com/apify/apify-core/issues/29471

Add ability to publish and un-publish tasks via [new API ](https://github.com/apify/apify-core/pull/29623)

Blocked by https://github.com/apify/apify-core/pull/29623

### DIFF
--- a/src/resource_clients/task.ts
+++ b/src/resource_clients/task.ts
@@ -68,49 +68,33 @@ export class TaskClient extends ResourceClient {
     }
 
     /**
-     * Publishes the task on its public landing page.
+     * Publishes the task on its public landing page, by setting `isPublic` through
+     * {@apilink TaskClient.update}.
      *
      * The task's Actor must be public and the task must have its public display configuration
-     * (`publicConfig`) set up via {@apilink TaskClient.update}. Requires write permission to
-     * both the task and its Actor.
-     *
-     * Publishing an already published task is a no-op that returns the current state, so this
-     * call is safe to retry (unlike {@apilink TaskClient.unpublish}).
+     * (`publicConfig`) set up first. Requires write permission to both the task and its Actor.
+     * Publishing an already published task does nothing.
      *
      * @returns The task object.
-     * @see https://docs.apify.com/api/v2/actor-task-publish-post
+     * @see https://docs.apify.com/api/v2/actor-task-put
      */
     async publish(): Promise<Task> {
-        const response = await this.httpClient.call({
-            url: this._url('publish'),
-            method: 'POST',
-            params: this._params(),
-        });
-
-        return cast(parseDateFields(pluckData(response.data)));
+        return this.update({ isPublic: true });
     }
 
     /**
-     * Unpublishes the task from its public landing page.
+     * Unpublishes the task from its public landing page, by setting `isPublic` through
+     * {@apilink TaskClient.update}.
      *
      * The public display configuration (`publicConfig`) is preserved, so the task can be
-     * published again without re-entering it. Requires write permission to both the task
-     * and its Actor.
-     *
-     * Unlike {@apilink TaskClient.publish}, this call is not idempotent: unpublishing a task
-     * that is not currently published throws an `ApifyApiError` (`cannot-unpublish-actor-task`).
+     * published again without re-entering it. Requires write permission to both the task and its
+     * Actor. Unpublishing a task that is not published does nothing.
      *
      * @returns The task object.
-     * @see https://docs.apify.com/api/v2/actor-task-unpublish-post
+     * @see https://docs.apify.com/api/v2/actor-task-put
      */
     async unpublish(): Promise<Task> {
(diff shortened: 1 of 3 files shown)


Janjiran:
> ⚠️ There are broken links in the documentation.
> 
> See more at https://github.com/apify/apify-client-js/actions/runs/30789669108#summary-91610366020

Waiting for https://github.com/apify/apify-docs/pull/2840

src/resource_clients/task.ts
```diff
@@ -293,6 +331,24 @@ export interface Task {
...
+ * {@apilink TaskClient.unpublish} to change the publication state.
+ */
+export interface TaskPublicConfig {
+    publishedAt: Date | null;
+    seoTitle?: string | null;
+    seoDescription?: string | null;
+    categorization?: string | null;
+    inputSchemaFields?: string[] | null;
+    datasetName?: string | null;
+    datasetView?: string | null;
```
katzino, line 348: Important: You're missing `highlightedFields?: string[] | null` as far as I remember

src/resource_clients/task.ts
```diff
@@ -67,6 +67,44 @@ export class TaskClient extends ResourceClient {
...
+        return cast(parseDateFields(pluckData(response.data)));
+    }
+
+    /**
+     * Unpublishes the task from its public landing page.
+     *
+     * The public display configuration (`publicConfig`) is preserved, so the task can be
+     * published again without re-entering it.
+     *
+     * @returns The task object.
```
katzino, line 93: Suggestion: Worth documenting that `unpublish()` is not idempotent while `publish()` is — the API throws `cannotUnpublishActorTask` when the task isn't published whereas publishing an already-published task returns 200.

src/resource_clients/task.ts
```diff
@@ -67,6 +67,44 @@ export class TaskClient extends ResourceClient {
         return this._update(newFields);
     }
 
+    /**
+     * Publishes the task on its public landing page.
+     *
+     * The task's Actor must be public and the task must have its public display configuration
+     * (`publicConfig`) set up via {@apilink TaskClient.update}.
```
katzino: Suggestion: I would also mention that you need write access to the Actor.

src/resource_clients/task.ts
```diff
@@ -317,7 +373,7 @@ export interface TaskOptions {
  */
 export type TaskUpdateData = Partial<
     Pick<Task, 'name' | 'title' | 'description' | 'options' | 'input' | 'actorStandby'>
->;
+> & { publicConfig?: Omit<TaskPublicConfig, 'publishedAt'> | null };
```
katzino: Question: `| null` suggests that you can nullify this `publicConfig` object which you technically can't (we just ignore it while updating the rest) so I would probably drop it here and also in `apify-core`, WDYT?

src/resource_clients/task.ts
Janjiran: It's stale no? I don't think it's used anywhere and if we decide to use it, the I would add it to public facing API, wdyt?

katzino: Hmm, good catch I guess we kept duplicated field to `inputSchemaFields`. No idea how that happened, we should remove it then.

src/resource_clients/task.ts
```diff
@@ -317,7 +381,7 @@ export interface TaskOptions {
  */
 export type TaskUpdateData = Partial<
```
katzino, line 371: Note: I commented on it at https://github.com/apify/apify-core/pull/29623#discussion_r3682214152 - `TaskCreateData` extends this type that means create method takes `publicConfig` but the endpoint in `apify-core` currently won't set it. Should be addressed there, not a blocker in this PR.

katzino (approved):
LGTM, two comments, mostly informative ones.

test/tasks.test.ts
```diff
@@ -97,6 +97,30 @@ describe('Task methods', () => {
...
+            const res = await client.task(taskId).publish();
+            expect(res.id).toEqual('publish-task');
+            validateRequest({ query: {}, params: { taskId } });
+
+            const browserRes = await page.evaluate((id) => client.task(id).publish(), taskId);
+            expect(browserRes).toEqual(res);
+            validateRequest({ query: {}, params: { taskId } });
+        });
+
+        test('unpublish() works', async () => {
```
B4nan, line 112: is this dependent on the previous test? Ideally, tests should work in isolation, if they are dependent, maybe just merge them to a single test case

B4nan (approved):

src/resource_clients/task.ts
Janjiran: https://github.com/apify/apify-core/issues/29803

test/tasks.test.ts
Janjiran: They are independent :)

PR #991, opened 2026-08-03
docs: update CONTRIBUTING.md to match actual tooling

B4nan:
CONTRIBUTING.md still described the setup from before the oxc migration, so this brings it in line with the actual scripts in `package.json`:

- npm commands are now pnpm (the repo uses `packageManager: pnpm@10.33.4`), including the `npm 10+` prerequisite
- the lint and format scripts are described as oxlint (`--type-aware`) and oxfmt instead of ESLint and Prettier
- the Code Style section points to `.oxfmtrc.json` and `oxlint.config.ts` instead of the removed `.prettierrc` and `eslint.config.mjs`
- the `npx vitest` examples now use `pnpm vitest`

This only touches CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ Thank you for your interest in contributing to the official JavaScript/TypeScrip
 ### Prerequisites
 
 - Node.js 18+ (LTS recommended)
-- npm 10+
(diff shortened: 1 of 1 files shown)


vdusek (approved):
lgtm

PR #995, opened 2026-08-10
test: add integration test suite

vdusek:
Port of the integration test suite from the Python API client.

Adds an integration test suite that executes the client against the live Apify API. This is the same approach as in the Python API client: real API calls with a test user token, resources created under unique names and cleaned up afterwards, and eventual consistency handled by polling helpers rather than sleeps or retries.

Merge this **before** #985, #986 and the rest of the v3 work, so those changes have some end-to-end test coverage.

## Coverage

196 tests over Actors, Actor versions, Actor env vars, builds, runs, logs, tasks, schedules, webhooks, webhook dispatches, datasets, key-value stores, request queues, the store, and users.

## Test tiers

`vitest.config.mts` now defines two projects, so the existing unit tests stay fast and offline:

- `pnpm test` - unit tests only (unchanged behavior)
- `pnpm test:integration` - integration tier only
- `pnpm test:all` - both

## Credentials

The tier needs `APIFY_TEST_USER_API_TOKEN`.

## CI

A new `integration_tests` job in `check.yaml` runs the tier on Node 26. It is skipped for fork PRs, where repository secrets are unavailable, and can be triggered on demand via `workflow_dispatch`.

*✍️ Drafted by Claude Code*


B4nan (approved):
LGTM, nice suite.

One thing unrelated to the diff: the `v3` branch looks outdated compared to `master`. For example, CONTRIBUTING.md here still documents the npm commands, while on `master` it already uses pnpm. Probably worth syncing `v3` with `master`.

PR #1011, opened 2026-08-26
fix: prevent path traversal by sanitizing URLs

### LINKED ISSUE #1005: URL-encode record keys and request IDs to prevent path traversal
## Problem

The client interpolates record keys, request IDs, and resource IDs directly into request URLs. Values containing `/`, `?`, `#`, or dot segments can therefore change the target path, query, or fragment.

For example, `deleteRecord('..')` resolves to the key-value store root and sends `DELETE /v2/key-value-stores/{storeId}`, potentially deleting the entire store. Other crafted values can target a different endpoint using the caller's token or silently access the wrong record.

This is the JS equivalent of apify/apify-client-python#1025, which has already been fixed.

## Cause

`_url()` uses plain string interpolation:

```ts
protected _url(path?: string): string {
    return path ? `${this.url}/${path}` : this.url;
}
```

Affected values are interpolated in:

- `key_value_store.ts`: record operations
- `request_queue.ts`: request and lock operations
- `api_client.ts`: resource IDs

Additionally, `_toSafeId()` uses `id.replace('/', '~')`, which replaces only the first slash.

## Expected fix

- Encode each dynamic path segment with `encodeURIComponent()` at its call site. Do not encode the complete path, because paths such as `requests/batch` contain intentional separators.
- Reject `''`, `.` and `..`; URL parsers resolve dot segments even when percent-encoded.
- Keep record signatures based on the raw key, as the server verifies the decoded route parameter.
- Make `_toSafeId()` replace every slash, matching the Python client.

barjin:
Swaps string interpolation in URL building for sanitized composition. 

Related to https://github.com/apify/apify-client-python/pull/1025
Closes #1005


vdusek (commented):
As we discussed on team sync, maybe this should target v3? cc @B4nan

B4nan:
Yeah, let's retarget. If this only popped during internal audits and nobody is waiting for the fix, it's better to stay safe.

janbuchar (approved):
Looks good to me code-wise.

vdusek (changes requested):
Requesting changes for retarget to v3

vdusek (approved):
LGTM

docs/public-api/apify-client.api.md
```diff
@@ -743,6 +744,12 @@ interface ApifyResponse<T = any> extends AxiosResponse<T> {
     config: ApifyRequestConfig & InternalAxiosRequestConfig;
 }
 
+// @public
+export class ArgumentValidationError extends Error {
```
barjin, line 748: These changes are from the previous commits to `v3`, the checks were just failing without them.

PR #1013, opened 2026-08-26
chore: support next major canaries and RC releases in the publish workflow

B4nan:
Adds publishing for the next major line to the `v3` branch, same setup as in apify-sdk-js (apify/apify-sdk-js#698) and crawlee (apify/crawlee#4009).

Automatic canaries: the new `pre_release_next_major.yaml` workflow publishes `3.0.0-beta.N` under the `next-v3` dist-tag on every push to `v3`. Unlike the pre-releases from master, it produces no changelog and no commit; `package.json` stays on the current 2.x version. `check.yaml` now also runs on pushes to `v3` so the publish can wait for the checks.

Manual RC releases: dispatching `publish_to_npm.yaml` with `tag: rc` publishes the next free `3.0.0-rc.N` under the `rc` dist-tag and pushes a `v3.0.0-rc.N` git tag pointing at the released commit. Same as the canaries, no changelog and nothing committed to the branch; the version bump lives only on npm.

`before-beta-release.js` now takes the dist-tag as an argument and derives the version from it: `next-v3` publishes betas of the major named by the tag, `rc` publishes release candidates of the next major, and the plain `beta` flow from master keeps its current behavior. The prerelease counter is scoped per preid (`beta.` vs `rc.`) so the two series advance independently, and the scan no longer matches unrelated versions through the old `startsWith` check (`2.25.1` vs `2.25.10`).

Verified the version math against live npm data: `rc` computes 3.0.0-rc.0, `next-v3` computes 3.0.0-beta.0, `beta` computes 2.25.1-beta.8 (current beta is 2.25.1-beta.7).

Everything lives on the `v3` branch only: push-triggered workflows run from the pushed branch's files, `execute-workflow` dispatches on the caller's ref, and manual dispatches use the file from the branch selected in the run dropdown. `master` needs no changes.

### DIFF
--- a/.github/scripts/before-beta-release.js
+++ b/.github/scripts/before-beta-release.js
@@ -9,28 +9,55 @@ const PKG_JSON_PATH = path.join(__dirname, '..', '..', 'package.json');
 const pkgJson = require(PKG_JSON_PATH);
 
 const PACKAGE_NAME = pkgJson.name;
-const VERSION = pkgJson.version;
 
-const nextVersion = addBetaSuffixToVersion(VERSION);
+// The npm dist-tag we publish to is passed as the first CLI argument. A tag
+// ending in `v<major>` (e.g. `next-v3`) marks a pre-major release line:
+// package.json stays on the current major (2.x) and we publish betas of the
+// *next* major (3.0.0-beta.N) under that tag. (The tag isn't a bare `v3`
+// because npm refuses dist-tags that parse as a semver range.)
+// The `rc` tag works the same way but publishes release candidates of the
+// next major (e.g. 3.0.0-rc.N).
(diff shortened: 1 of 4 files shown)


vdusek (approved):
If I understand it correctly, with this setup, I won't have to be testing packages accross repos with this git references setup - https://github.com/apify/apify-sdk-js/blob/3703ed859f121b70a4e319107b52dcaeaa1ce44e/package.json#L85, right?

Otherwise, LGTM

B4nan:
Yep, every commit to `v3` branch will ship a new major dev version (based on latest stable, so more commits there will keep bumping the `v3.0.0-beta.x`), which you will use for cross-repo integration tests.

PR #1015, opened 2026-08-27
refactor!: convert to native ESM

vdusek:
Closes:  #1025 

Mirrors [apify/crawlee@cc01b18](https://github.com/apify/crawlee/commit/cc01b18ee1fa026a2c22ddb859c82f49a98db051) and [apify/apify-sdk-js@a381728](https://github.com/apify/apify-sdk-js/commit/a381728530c42c25d975348b77c46789ca9c2b2f), which did the same for Crawlee v4 and for the SDK on its `v4` branch.

## What changed

- `"type": "module"`, and the dual CJS+ESM build is gone: no more `dist/index.mjs`, no `gen-esm-wrapper` postbuild step, and the `import`/`require` export conditions collapse into `types` + `browser` + `default`.
- Every relative import in `src/` and `test/` gets its `.js` extension, which is the bulk of the diff. It follows from `"type": "module"` alone: the files now emit as ESM, and ESM has no extension guessing or directory-index lookup. `tsc` never rewrites specifiers, so dropping them type-checks clean and then fails at runtime with `ERR_MODULE_NOT_FOUND`. crawlee v4 and apify-sdk-js `v4` do the same.
- `require('../package.json')` in `getVersionData()` becomes a JSON import attribute, with `resolveJsonModule` off so the emit doesn't sink into `dist/src`.
- `__dirname` becomes `import.meta.dirname` in the test helpers and the Vitest config.
- tsconfig pins `module`/`moduleResolution` to `NodeNext` and `target` to `ESNext`, matching crawlee v4 and the SDK.
- `test/tsconfig.json` maps `apify-client` to `../src/index.ts` rather than the `../src` directory. `nodenext` has no directory-index lookup, so the directory form type-checked the tests against `dist`.

## The browser bundle is now an ES module

`dist/bundle.js` keeps its name and path, but rsbuild now emits it as an ES module with named exports instead of UMD. The `Apify` global is gone: load it with `<script type="module">` and `import { ApifyClient } from '.../bundle.js'`. Under `"type": "module"` Node parses a `.js` file as ESM anyway, and the UMD file in that spot broke `pnpm test:bundling` on all four bundlers. The `apify-client/browser` subpath is unchanged.

## Breaking change

The client is pure ESM and ships no CommonJS build. `require('apify-client')` still resolves on Node.js 22.12 and newer, which loads an ES module through `require()` directly. Anywhere else, CommonJS consumers have to switch to `import` or a dynamic `import()`. Documented in the v3 upgrading guide, along with the bundle format change.

*✍️ Drafted by Claude Code*


docs/04_upgrading/upgrading_v3.md
```diff
@@ -9,6 +9,23 @@ import ApiLink from '@theme/ApiLink';
 
 This page summarizes the breaking changes when upgrading from v2 to v3 of `apify-client`.
 
+## The package is now pure ESM
+
+`apify-client` ships as an ES module. The CommonJS build is gone, along with the `dist/index.mjs` wrapper and the `require` condition in `exports`, so `import` is the supported way to load the client.
+
+```diff
+- const { ApifyClient } = require('apify-client'); // v2
++ import { ApifyClient } from 'apify-client';     // v3
```
B4nan: ocd


```suggestion
+ import { ApifyClient } from 'apify-client';      // v3
```

docs/04_upgrading/upgrading_v3.md
```diff
@@ -9,6 +9,23 @@ import ApiLink from '@theme/ApiLink';
...
+```
+
+A CommonJS project can still `require('apify-client')` on Node.js 22.12 or newer, which loads an ES module through `require()` directly. Anywhere without that support, switch to `import`, either by adding `"type": "module"` to your `package.json` or through a dynamic `import()`:
```
B4nan: this sentence is a bit confusing
- require(esm) works in older versions than node 22.12 too, it was backported to 20.x line
- we support only node 22+ anyway, dynamic import workaround is surely not needed to use this package

i would maybe just drop it, we want to require node 22+ just like all the other packages, then this is non issue

docs/04_upgrading/upgrading_v3.md
```diff
@@ -9,6 +9,23 @@ import ApiLink from '@theme/ApiLink';
...
+
+```js
+const { ApifyClient } = await import('apify-client');
+```
+
+The browser bundle moved from `dist/bundle.js` to `dist/bundle.cjs`, because a `.js` file inside an ESM package is parsed as an ES module, and the bundle is UMD. The `apify-client/browser` subpath is unchanged, so only code that pointed at the file path itself needs updating. For details, see [Bundled environments](../02_concepts/05_bundled-environments.md).
```
B4nan: huh, browser bundle should be ESM too, It makes no sense to publish a browser bundle as CJS, browsers never understood CJS, that is a nodejs thing

i get the reasoning about UMD, but i feel like this should just be ESM too with the new major

cc @vladfrangu you might know better?

package.json
```diff
@@ -5,19 +5,18 @@
     "engines": {
         "node": ">=22.0.0"
```
B4nan, line 6: now the note about node 22.12 makes more sense, i didnt realize it wasnt part of node 22.0...

docs/04_upgrading/upgrading_v3.md
vdusek: The whole paragraph was rephrased, and the dynamic import workaround was dropped. Ready for a recheck.

docs/04_upgrading/upgrading_v3.md
vdusek: the browser bundle should be switched to ESM too, you can recheck it

B4nan (approved):
looks good to me, left some tiny fixes found during my review

Issue #522, opened 2024-02-26
`parseDateFields` skips date fields nested deeper than 3 levels


tobice:
[`parseDateFields`](https://github.com/apify/apify-client-js/blob/980d958ee5e6bf8610f7ce08fb282e126c68fa82/src/utils.ts#L62) traverses the provided JSON (used for Apify API responses) and converts certain fields to a `Date` object. The traversion depth has been historically limited to 3, possibly to avoid cyclic dependencies and improve performance on large responses.

This limitation is not communicated in any way, which might lead to weird surprises. We tried adding a `log.warning` but that led to log pollution as there are objects with arbitrary structure that went over the depth limit (most likely user data, data sets). Context [here](https://apifier.slack.com/archives/CGZSN9DQC/p1708707905914559).

Possible solutions:
- Remove the limitation
- Stop parsing objects with arbitrary structure (user data, data sets)

Both are potential breaking changes.

This is currently not causing any issues and can be treated with low priority.

drobnikj:
Let's check with @B4nan, but I guess this is low-prio for now.

B4nan:
Yeah, this can wait, we just wanted to have this tracked. It sounds like a good first issue for some new joiners.

vdusek:
resolved by https://github.com/apify/apify-client-js/pull/1060

PR #1016, opened 2026-08-27
feat!: validate API responses with zod schemas generated from the OpenAPI spec

vdusek:
Closes:  #1027 

## Description

This PR adds runtime validation of API responses using [Zod](https://zod.dev/) schemas generated from the OpenAPI specification, similar to how the Python API client validates responses with [Pydantic](https://pydantic.dev/).

It follows up on #985, which introduced generated TypeScript types via [openapi-typescript](https://openapi-ts.dev/).

## Schema generation

- `pnpm generate:models` now generates both:
  - `src/generated/api.ts` — TypeScript types
  - `src/generated/schemas.ts` — Zod schemas
- The nightly `regenerate_models.yaml` workflow regenerates both.
  - Schemas are generated by a small custom OpenAPI-to-Zod emitter in `scripts/schema_emitter.mts`.
- The generated schemas intentionally:
  - use loose objects, so unknown response fields are preserved,
  - treat string enums as open, so unknown enum values are accepted,
  - use `z.date()` for `date-time`, because `parseDateFields()` runs before validation,
  - fail generation on unsupported JSON Schema keywords instead of silently ignoring them.

I also considered existing generators - [hey-api](https://heyapi.dev/), [orval](https://orval.dev/), [kubb](https://kubb.dev/). None of them supports all the behaviors above natively, so the custom code would not disappear; it would move into resolvers or post-processing. Since our specification uses only a small subset of JSON Schema and all 240 generated schemas are cross-checked at compile time against the independently generated TypeScript types, keeping the emitter local seems like a better solution, and thanks to that to also avoid another dependency.

## Validation

- `parseResponse()` now unwraps the API response, parses date fields, and validates the result. All resource methods in the base clients go through it.
- Invalid responses throw a new `ResponseValidationError`, which includes:
  - the request method and URL,
  - validation issues,
  - the original cause,
  - a message identifying the offending fields.
- The API deviations found along the way were fixed in the specification itself (apify/apify-docs#2932), so there are no hand-written schema overrides now.

## Tests

- The mock server now uses spec-shaped fixtures generated from OpenAPI examples (`test/mock_server/fixtures.ts`), with their validity checked in `fixtures.test.ts`.
- Tests that intentionally use synthetic responses, such as pagination, timeout, and URL-encoding tests, mock the validation step.
- The integration suite remains the main check for real API/spec drift and passes on this branch.

## Breaking changes

- Responses that do not match the OpenAPI specification now throw `ResponseValidationError`.
- `ScheduleClient.getLog()` now returns `ScheduleInvoked[]` instead of `string`, matching the actual API response.
- `TaskPublicConfig` now follows the specification:
  - `publishedAt` is optional and read-only,
  - `categorization` is removed.

All breaking changes are documented in the v3 upgrading guide.

## Bundle size

The browser bundle grows from **288 kB** to **327 kB** due to the generated schemas. The limit in `rsbuild.config.ts` is increased to **360 kB**.

*✍️ Drafted by Claude Code*


B4nan:
The date parsing changes reminded me of #522, we should open this topic on slack in some public channel and agree on the direction.

vdusek:
> The date parsing changes reminded me of https://github.com/apify/apify-client-js/issues/522, we should open this topic on slack in some public channel and agree on the direction.

Sure, but I guess it does not block this PR, right? And we can do it later as a separate PR.

B4nan:
Yeah, that should be resolved separately, including an entry in the upgrading guide.

src/generated/schemas.ts
barjin, line 1: Importing this file takes up to 200 ms (156 ms median). That basically wipes out all our import perf improvements in Crawlee. Any chance we could improve this somehow? (somehow lazy-build the schemas, for example)? 

Image

(The first measurement is importing `dist/index.js` in `v3`, the second is the current branch, the third is importing this schema file alone)

src/response_validation_error.ts
```diff
@@ -0,0 +1,40 @@
...
+    /** HTTP method of the request whose response failed validation, upper-cased. */
+    readonly method: string;
+
+    /** URL of the request whose response failed validation, without its query string. */
+    readonly url: string;
+
+    constructor(error: z.ZodError, value: unknown, request: { method: string; url: string }) {
+        const method = request.method.toUpperCase();
+        super(
+            `Response from ${method} ${request.url} does not match the API schema:\n${formatZodError(error, value)}`,
```
janbuchar: If we're asking users to report this in the error handling docs, perhaps we should also do so here?

scripts/generate_schemas.mts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Generates `src/generated/schemas.ts` -- a zod schema per `components.schemas` entry -- from the OpenAPI
+ * specification downloaded into `tmp/openapi.json`.
+ *
+ * Run via `pnpm generate:models`, which downloads the specification first, generates the types next to these
+ * schemas, and records the specification version afterwards. Requires a Node release that strips TypeScript
+ * syntax natively (>=22.18 or >=23.6).
```
janbuchar: Is this comment necessary?

janbuchar (approved):
Thanks, other than my two comments, I didn't notice anything suspicious.

scripts/generate_schemas.mts
vdusek: removed

src/generated/schemas.ts
vdusek: The zod schemas are now constructed lazily, instead of at import time. See 0f38e24503fd77f1c715e2afba4b803c625ea71f. Could you re-check please?

B4nan: If it works well, let's do the same in the other places (SDK, crawlee) right ahead.

src/lazy_schema.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Wraps the construction of a zod schema so it runs on the first call and only once.
+ *
+ * The generated response schemas are exported this way: a couple of hundred `z.looseObject()` calls at import time
+ * would cost more than importing zod itself, so each schema is built by the first response validated against it and
+ * kept for the rest of the process.
+ */
+export function lazySchema<T>(build: () => T): () => T {
```
B4nan, line 8: lets move this to apify-shared

https://github.com/apify/apify-shared-js/tree/master/packages/validations

(but first let's wait for @barjin to retest the perf effects)

B4nan: also cc @vladfrangu, what do you think about this approach?

vladfrangu: its confusing because yesterday we benchmarked this in crawlee v4 and creating the schemas lazily didn't improve anything much import-time wise... 👀 (in BasicCrawler, creating the schema lazily did shave off like 10ms off import times, but importing zod added much more time anyways)

Me and Jindra are benchmarking now and, while this PR does shave off ms from importing the whole apify-client, the move to zod added more (on my laptop, we went from ~70ms import time on master to 130ish on v3 to 100ish on this PR)

src/generated/schemas.ts
barjin: Thank you @vdusek , it seems to have shaven off approx. 50 ms on my Linux machine (making the measurements less noisy, too). 

import_times_chart

The bump from the current `v2` branch is striking, but I suppose it's not all just the `zod` schemas in that case.

Imo the current tip of this PR branch is comparable to `v3` (perf-wise) :+1: Idk if extracting this helper is necessarily worth it. I don't recall any other Apify projects with such extensive schemas (at the module's top level).

barjin (approved):
lgtm, thank you @vdusek 👍

src/lazy_schema.ts
barjin: (see https://github.com/apify/apify-client-js/pull/1016#discussion_r3956419389 for a chart of :arrow_up: )

src/generated/schemas.ts
B4nan: could we import zod dynamically too? 

on the other hand, not sure if this can actually help, maybe with import time, but people will surely use something from the client, making this only a technicallity, not actual perf win.

if its comparable, lets just ship it now and deal with this later (or never)

B4nan (approved):

src/generated/schemas.ts
barjin: Agreed, I'd still like to look into the v2 -> v3 regression, but as far as this PR goes, there isn't much more we can do.

vdusek:
Thanks everyone. I'm merging it. As mentioned in discussions, we can do deeper v2 -> v3 benchmarking and testing separately (any further performance optimizations, extracting reusable parts into `apify-shared` where it makes sense, and more).

src/lazy_schema.ts
B4nan: we should also explore using zod/mini

https://zod.dev/packages/mini

PR #1040, opened 2026-09-08
fix!: type the Actor run input on ActorClient and RunClient

### LINKED ISSUE #818: Align `ActorClient.start()` and `call()` input typing with `TaskClient`
`ActorClient`'s methods `start` and `call` have as a first parameter `input?: unknown`. The same methods in `TaskClient` have `input?: Dictionary`. This is inconsistent, and allows passing an input with the wrong type when using `ActorClient`.

vdusek:
`ActorClient.start()`, `call()`, `validateInput()` and `RunClient.metamorph()` took `input` as `unknown`, so anything compiled, including values the client can't send. They now take `ActorInput`, an alias for `object`: a JSON-serializable object or array. A string, a number, a boolean, `null` and values typed `unknown` stop compiling. Nothing changes at runtime. `metamorph()`'s `input` becomes optional too.

The raw `string` body that v2 accepted alongside `contentType` is left out on purpose, as non-JSON Actor input is being sunset. The `contentType` option itself stays.

The issue asked for `Dictionary`, the type `TaskClient` uses. That rejects anything typed by your own `interface`, which gets no index signature. Task input keeps `Dictionary` because its overrides merge into the input saved on the task.

The `ActorStandby` fields from the issue's second comment are already on `v3` via #985. `apify-sdk-js` forwards `input?: unknown` into this client, so it needs the same retyping before it bumps.

Closes #818

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -555,8 +555,8 @@ export type ActorUpdateOptions = Partial<
 >;
 
 /**
- * Input for an Actor run, as taken by {@link ActorClient.start}, {@link ActorClient.call} and
- * {@link ActorClient.validateInput}.
+ * Input for an Actor run, as taken by {@link ActorClient.start}, {@link ActorClient.call},
+ * {@link ActorClient.validateInput} and {@link RunClient.metamorph}.
  *
  * Without `contentType` in the options, the input is serialized to JSON, so it is an object or an array.
  * With `contentType`, the input is sent as the request body as-is, so it is a `string` or a `Buffer`.
--- a/src/resource_clients/actor.ts
+++ b/src/resource_clients/actor.ts
@@ -161,8 +161,7 @@ export class ActorClient extends ResourceClient {
      * asynchronously and this method returns immediately without waiting for completion.
      * Use the {@link call} method if you want to wait for the Actor to finish.
      *
-     * @param input - Input for the Actor. A JSON-serializable object or array, or a `string` or `Buffer` sent
-     *                as-is when `contentType` is specified in options. Omit it to run the Actor without input.
+     * @param input - Input for the Actor, serialized to JSON. Omit it to run the Actor without input.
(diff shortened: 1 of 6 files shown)


src/resource_clients/actor.ts
```diff
@@ -161,8 +161,8 @@ export class ActorClient extends ResourceClient {
      * asynchronously and this method returns immediately without waiting for completion.
      * Use the {@link call} method if you want to wait for the Actor to finish.
      *
-     * @param input - Input for the Actor. Can be any JSON-serializable value (object, array, string, number).
-     *                If `contentType` is specified in options, input should be a string or Buffer.
+     * @param input - Input for the Actor. A JSON-serializable object or array, or a `string` or `Buffer` sent
```
janbuchar: What does a `string` input mean? Also, I was under the impression that non-JSON Actor inputs will be... sunset.

test/actors.test.ts
```diff
@@ -434,6 +434,28 @@ describe('Actor methods', () => {
...
         });
 
+        test('start(), call() and validateInput() take the same input type', () => {
+            // An interface gets no implicit index signature, so a `Record`-based input type rejects it.
+            interface UserInput {
+                url: string;
+            }
+
+            const actor = client.actor('some-id');
+            expectTypeOf(actor.start).parameter(0).toEqualTypeOf<ActorInput | undefined>();
```
janbuchar: Just my curiosity, is this testing the `ActorInput` type definition? Is there any value in that?

janbuchar (approved):
LGTM

test/actors.test.ts
vdusek: removed

src/resource_clients/actor.ts
vdusek: only JSON Actor inputs now

PR #1041, opened 2026-09-08
feat!: add ApifyApiError subclasses grouped by HTTP status

### LINKED ISSUE #709: Find a way to distinguish API error types in error handling code
- https://apify.slack.com/archives/CD0SF6KD4/p1750081551997999
- We probably want an `ApifyApiError` subclass for each error type
    - In `catch` blocks, we would just use `instanceof`
    - Or we can have typescript type guard functions instead
- It might make sense to generate this code

vdusek:
The client now throws the `ApifyApiError` subclass matching the response's HTTP status code, so a `catch` block can branch on `instanceof` instead of comparing `statusCode` numbers or `type` strings. The split follows the status code because `type` is per-endpoint and has hundreds of values. Same approach and same class names as the Python client in apify/apify-client-python#737.

| Status | Class |
| --- | --- |
| 400 | `InvalidRequestError` |
| 401 | `UnauthorizedError` |
| 403 | `ForbiddenError` |
| 404 | `NotFoundError` |
| 409 | `ConflictError` |
| 429 | `RateLimitError` |
| 5xx | `ServerError` |

Any other status stays a plain `ApifyApiError`, and every subclass extends it, so existing `instanceof ApifyApiError` checks keep working. `HttpClient` builds the error through a hidden `ApifyApiError.fromResponse()` factory. The constructor is unchanged.

`type` is now typed as `LiteralUnion<ApifyApiErrorType, string>`, with `ApifyApiErrorType` coming from the spec's `ErrorType` enum. Editors autocomplete `err.type === 'actor-memory-limit-exceeded'`, which is what the Slack thread asked for, and any string the API returns still type-checks.

## Breaking changes

- `error.name` now holds the subclass name, so a printed stack starts with `NotFoundError: ...` instead of `ApifyApiError: ...`.
- `catchNotFoundOrThrow` checks `instanceof NotFoundError`, so `get()`-style methods swallow every 404 whatever its `type`. Before, only `record-not-found`, `record-or-token-not-found` and HEAD requests were swallowed. The Python client made the same change, and the v3 upgrading guide covers it.
- The same helper backs `waitForFinish()` and `call()`, which read a swallowed 404 as "the run isn't visible yet". A 404 that used to throw right away now keeps them polling until `waitSecs` runs out, and they end up throwing a generic error instead of the `ApifyApiError`. A token revoked mid-wait lands there.

Closes #709

*✍️ Drafted by Claude Code*

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,9 @@ Besides greatly simplifying the process of querying the Apify API, the client pr
 Based on the endpoint, the client automatically extracts the relevant data and returns it in the
 expected format. Date strings are automatically converted to `Date` objects. For exceptions,
 we throw an `ApifyApiError`, which wraps the plain JSON errors returned by API and enriches
-them with other context for easier debugging.
+them with other context for easier debugging. The error is an instance of the subclass matching the
+HTTP status code, such as `NotFoundError` or `RateLimitError`, so a `catch` block can tell them apart
+with `instanceof`.
 
 ### Retries with exponential backoff
 
--- a/docs/02_concepts/02_error-handling.md
+++ b/docs/02_concepts/02_error-handling.md
@@ -69,7 +69,7 @@ try {
     await client.actor('my-actor').call({ url: 'https://example.com' }, { memory: 32768 });
 } catch (error) {
     if (error instanceof ApifyApiError && error.type === 'actor-memory-limit-exceeded') {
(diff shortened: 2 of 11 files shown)


barjin (approved):
lgtm, thank you @vdusek !

PR #1042, opened 2026-09-08
fix!: throw on ambiguous 404s instead of resolving to undefined

### LINKED ISSUE #1030: Throw on ambiguous 404s instead of returning `undefined`
Every 404 is swallowed into `undefined`. Where a 404 can mean either the parent or the sub-resource is missing, that hides a real error: `run.dataset()` / `keyValueStore()` / `requestQueue()` / `log()`, `dataset.getStatistics()`, `user.monthlyUsage()` and `limits()`, `schedule.getLog()`, `task.getInput()`, `webhook.test()`. A direct `get(id)` / `delete(id)` keeps returning `undefined`.

Breaking — the affected return types drop `| undefined`. Needs the error subclasses from #709 first.

Python: apify/apify-client-python#755, on top of apify/apify-client-python#737.

*✍️ Drafted by Claude Code*

vdusek:
A 404 collapsed into `undefined` everywhere, including where it can't be pinned to one resource: `run.dataset().get()` couldn't tell a missing run from a missing dataset. Those calls now throw.

- Chained clients without an ID (`run.dataset()`, `run.keyValueStore()`, `run.requestQueue()`, `run.log()`, `build.log()`) throw from `get()`, `delete()`, `log().get()` and `log().stream()`. A new `catchNotFoundForResourceOrThrow(err, this.id)` keys on the ID, so ID-addressed clients still resolve to `undefined`.
- Fixed sub-paths throw and drop `| undefined`: `getStatistics()`, `monthlyUsage()`, `limits()`, `getLog()`, `getInput()`, `test()`.
- Unchanged: `getRecord()`, `getRequest()`, `recordExists()` and `lastRun()`.
- `version()`, `build()` and `envVar()` reject an empty string, which used to address the collection.
- `UserClient.get()` is now `Promise<User | undefined>`, matching what it always returned.
- `getStreamedLog()` on a missing run warns and stops.

Merging `v3` also moved `LogClient.stream()` under the same rule. #1041 made `catchNotFoundOrThrow()` a plain `instanceof NotFoundError` check, so a streamed 404 matches it and `client.log(id).stream()` resolves to `undefined` the way `get()` does. #1043 raised that flip as an open question; the unusable error body it reports is still open.

Mirrors apify/apify-client-python#755.

Closes #1030

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/resource_clients/run.ts
+++ b/src/resource_clients/run.ts
@@ -377,8 +377,8 @@ export class RunClient extends ResourceClient {
     /**
      * Returns a client for the default key-value store of this Actor run.
      *
-     * A 404 from this client throws an `ApifyApiError` rather than resolving to `undefined`, since the run itself
-     * may be what is missing.
+     * `get()` and `delete()` throw an `ApifyApiError` on a 404 rather than resolving to `undefined`, since the run
+     * itself may be what is missing. Record lookups such as `getRecord()` keep reading a 404 as a missing record.
      *
      * @returns A client for accessing the run's default key-value store
      * @see https://docs.apify.com/api/v2/actor-run-get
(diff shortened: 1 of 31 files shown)


src/resource_clients/run.ts
```diff
@@ -354,6 +354,9 @@ export class RunClient extends ResourceClient {
     /**
      * Returns a client for the default dataset of this Actor run.
      *
+     * A 404 from this client throws an `ApifyApiError` rather than resolving to `undefined`, since the run itself
```
janbuchar: This comment (and others like it) is arguing with the previous behavior, and the reader might not know what it was.

janbuchar (approved):

src/resource_clients/run.ts
vdusek: Updated, and Claude should remember that.

Issue #979, opened 2026-07-27
Pagination iterator mishandles server-side-filtered dataset pages


vdusek:
### Description

`DatasetClient.listItems()` used as an async iterable returns wrong results with `clean` / `skipEmpty` / `skipHidden`.

The API applies `offset`/`limit` to the raw item stream and filters **afterwards**, so a page can scan up to `limit` rows while returning fewer (or zero) items. `x-apify-pagination-count` reports the scanned window; the iterator ignores it. Two bugs in `_listPaginatedFromCallback`:

1. **Stops early** — `items.length > 0` (`src/base/api_client.ts:118`) ends iteration on a fully filtered page, even though later pages hold matching items.
2. **Yields duplicates** — `currentOffset += currentPage.items.length` (`src/base/api_client.ts:128`) advances by returned instead of scanned items, so the next page re-scans rows already yielded. The header is discarded at `src/resource_clients/dataset.ts:400`.

Cursor iterators (`listKeys`, `listRequests`, `paginateRequests`) are unaffected — the API never returns an empty page with a usable cursor.

Fixed in the Python client in apify/apify-client-python#964. A JS fix must read `x-apify-pagination-count` separately for bookkeeping, since `count` in `PaginatedList` is public and intentionally set to `data.length`.

API-side root cause: apify/apify-core#27324.

### Description from #1032 

`paginateRequests()` returns as soon as a page carries no items (`src/utils.ts:329`), before looking at `nextCursor`. With `filter: 'pending'` or `'locked'` the API can filter out a whole page while the cursor still points at more data, so the iterator truncates silently. `listKeys()` has the same stop condition (`src/resource_clients/key_value_store.ts:215`). Termination should rely on the cursor alone.

#979 states cursor iterators are unaffected — that does not hold for a server-side `filter`.

Python: apify/apify-client-python#964.

*✍️ Drafted by Claude Code*

vdusek:
resolved by https://github.com/apify/apify-client-js/pull/1044

PR #1044, opened 2026-09-08
fix!: keep pagination iterators advancing past fully filtered pages

vdusek:
### TLDR

- Mirrors apify/apify-client-python#964, which fixed the same bug in the Python client.
- The API-side root cause is described in apify/apify-core#27324.
- Closes: #979

### Description

The offset-based iterator stopped as soon as a page came back with no items, and advanced the offset by the number of items returned. Dataset filters (`clean`, `skipEmpty`, `skipHidden`) apply after `offset` and `limit`, so a page can scan up to `limit` rows while returning fewer, or none at all. `listItems()` used as an async iterable therefore ended early on a fully filtered page, and re-yielded rows an earlier page had already returned.

The iterator now paginates by the scanned count the API reports in `x-apify-pagination-count`, and by that number alone. `items.length` cannot stand in for it in either direction: filters leave it below the rows scanned and `unwind` can leave it above, so mixing the two would skip rows. `PaginatedList.count` stays `items.length`, since it is public and intentionally set that way; the scanned number travels with the page as a non-enumerable symbol property outside the public shape. Collection endpoints send no such header and fall back to `items.length`, so nothing about them changes.

The cursor iterators (`listKeys()`, `listRequests()`, `paginateRequests()`) are unchanged. The API never returns an empty page together with a cursor: the request queue listing hands out `nextCursor` only when it filled the whole page, and the key-value store listing derives `nextExclusiveStartKey` from the last key it returned.

### Test plan

Regression tests for the fully filtered page, the duplicated offset, and `unwind` returning more items than rows scanned.

### Breaking change

When `DatasetClient.listItems()` is consumed as an async iterable, `limit` caps the dataset rows the API scans, as it does on a single call, where v2 counted the items yielded. With `clean`, `skipEmpty` or `skipHidden` set, the loop can yield fewer than `limit` items, each of them once. v2 made up the difference with items it had already yielded, or stopped early on an empty page. With `unwind` the loop can yield more. The v3 upgrading guide covers it.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/base/api_client.ts
+++ b/src/base/api_client.ts
@@ -1,7 +1,7 @@
 import type { ApifyClient } from '../apify_client.js';
 import type { HttpClient } from '../http_client.js';
 import type { PaginatedResponse, PaginationOptions } from '../utils.js';
-import { toPath, toPathSegment } from '../utils.js';
+import { SCANNED_COUNT, toPath, toPathSegment } from '../utils.js';
 
 /** @private */
 export interface ApiClientOptions {
@@ -112,17 +112,24 @@ export abstract class ApiClient {
             limit: minForLimitParam(options.limit, chunkSize),
         } as T);
 
+        // A page can scan more items than it returns (see `SCANNED_COUNT`). The next offset and the stop condition
+        // follow the scanned number: advancing by `items.length` would re-scan items the previous page already
+        // returned, and stopping at an empty page would end the iteration in front of items a filter hid.
+        const scannedItems = (page: R): number =>
+            Math.max((page as { [SCANNED_COUNT]?: number })[SCANNED_COUNT] ?? 0, page.items.length);
+
         async function* asyncGenerator() {
             let currentPage = await paginatedListPromise;
             yield* currentPage.items;
             const offset = options.offset ?? 0;
             const limit = Math.min(options.limit || currentPage.total, currentPage.total);
 
-            let currentOffset = offset + currentPage.items.length;
-            let remainingItems = Math.min(currentPage.total - offset, limit) - currentPage.items.length;
+            let pageScanned = scannedItems(currentPage);
+            let currentOffset = offset + pageScanned;
+            let remainingItems = Math.min(currentPage.total - offset, limit) - pageScanned;
 
             while (
-                currentPage.items.length > 0 && // Continue only if at least some items were returned in the last page.
+                pageScanned > 0 && // Continue only if the last page scanned some items.
                 remainingItems > 0
             ) {
                 const newOptions = {
@@ -132,8 +139,9 @@ export abstract class ApiClient {
                 } as T;
                 currentPage = await getPaginatedList(newOptions);
                 yield* currentPage.items;
-                currentOffset += currentPage.items.length;
-                remainingItems -= currentPage.items.length;
+                pageScanned = scannedItems(currentPage);
+                currentOffset += pageScanned;
+                remainingItems -= pageScanned;
             }
         }
 
--- a/src/base/api_client.ts
+++ b/src/base/api_client.ts
@@ -112,11 +112,12 @@ export abstract class ApiClient {
             limit: minForLimitParam(options.limit, chunkSize),
         } as T);
 
-        // A page can scan more items than it returns (see `SCANNED_COUNT`). The next offset and the stop condition
-        // follow the scanned number: advancing by `items.length` would re-scan items the previous page already
-        // returned, and stopping at an empty page would end the iteration in front of items a filter hid.
-        const scannedItems = (page: R): number =>
-            Math.max((page as { [SCANNED_COUNT]?: number })[SCANNED_COUNT] ?? 0, page.items.length);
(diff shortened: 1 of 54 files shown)


src/resource_clients/key_value_store.ts
```diff
@@ -210,8 +210,8 @@ export class KeyValueStoreClient extends ResourceClient {
 
             let remainingItems = parsed.limit ? parsed.limit - currentPage.items.length : undefined;
 
+            // An empty page alone does not end the listing; only a missing next key does.
             while (
-                currentPage.items.length > 0 && // Continue only if at least some items were returned in the last page.
```
barjin, line 214: In KVS, we imo cannot get the "empty page" in the middle of a filtered list. 

The implementation just proxies the query [to the underlying S3 bucket](https://github.com/apify/apify-core/blob/2f46c1b5ea4306ed8b18a528e42307135206599f/src/api/src/lib/key_value_stores.ts#L815).

src/resource_clients/request_queue.ts
```diff
@@ -739,10 +739,9 @@ export class RequestQueueClient extends ResourceClient {
 
             let remainingItems = parsed.limit ? parsed.limit - currentPage.items.length : undefined;
 
-            // RQ API response does not indicate whether there are more requests left, so we have to try and in case
-            // of exhausting all requests we get response with empty items which ends the loop.
+            // A server-side `filter` can leave a page empty while `nextCursor` still points at more requests, so an
+            // empty page alone does not end the listing.
             while (
-                currentPage.items.length > 0 && // Continue only if at least some items were returned in the last page.
```
barjin, line 738: Same for RQ, this logic is already [handled on the backend](https://github.com/apify/apify-core/blob/2f46c1b5ea4306ed8b18a528e42307135206599f/src/api/src/routes/request_queues/requests.ts#L246), the page being empty means we're done-done.

src/base/api_client.ts
```diff
@@ -112,17 +112,24 @@ export abstract class ApiClient {
             limit: minForLimitParam(options.limit, chunkSize),
         } as T);
 
+        // A page can scan more items than it returns (see `SCANNED_COUNT`). The next offset and the stop condition
+        // follow the scanned number: advancing by `items.length` would re-scan items the previous page already
+        // returned, and stopping at an empty page would end the iteration in front of items a filter hid.
+        const scannedItems = (page: R): number =>
+            Math.max((page as { [SCANNED_COUNT]?: number })[SCANNED_COUNT] ?? 0, page.items.length);
```
barjin: I believe this is wrong, e.g., with `unwind=key`, `page.items.length` doesn't represent the scanned items count (you can `unwind` an array nested in the dataset item; that way, `page.items` will be the array items). We could accidentally skip parts of the dataset.

If I'm reading the API implementation correctly, it seems the `x-apify-pagination-count` header represents the actual scanned dataset item count (before filtering). This imo means we should paginate based solely on that and not `page.items`.

barjin (commented):
Thank you @vdusek ,

I have a few findings ⬇️ , the first two are just notes (perhaps these changes are not necessary), the last one imo a bit more actionable.

src/resource_clients/key_value_store.ts
vdusek: Nice, thanks. I'll review how this is done in Python.

src/resource_clients/request_queue.ts
vdusek: same answer as for https://github.com/apify/apify-client-js/pull/1044#discussion_r3977716039

src/base/api_client.ts
vdusek: Thanks. And again, I'll verify this in the Python API Client.

src/base/api_client.ts
```diff
@@ -112,17 +112,25 @@ export abstract class ApiClient {
...
             let currentPage = await paginatedListPromise;
             yield* currentPage.items;
             const offset = options.offset ?? 0;
             const limit = Math.min(options.limit || currentPage.total, currentPage.total);
 
-            let currentOffset = offset + currentPage.items.length;
-            let remainingItems = Math.min(currentPage.total - offset, limit) - currentPage.items.length;
+            let pageScanned = scannedRows(currentPage);
+            let currentOffset = offset + pageScanned;
+            let remainingItems = Math.min(currentPage.total - offset, limit) - pageScanned;
```
barjin, line 130: This seems to be a breaking change - with `limit` option set, we might get fewer items than before, since we're now including the skipped items (subtracting `pageScanned` instead of `currentPage.items.length`).

If this is intentional (it feels like it, given the direction of the other changes), I think it's worth a note in the upgrading guide.

barjin (approved):
Thank you @vdusek , one last nit ⬇️

src/base/api_client.ts
vdusek: PR title is updated, and a short section to the upgrade guide is added. Thank you.

PR #1045, opened 2026-09-08
feat!: normalize URL fields in responses

### LINKED ISSUE #1028: Normalize URL fields in responses
URL fields in responses are plain strings that pass through untouched — the generated zod schemas deliberately do not enforce `format: uri`. The Python client types them as `pydantic.AnyUrl`, which normalizes during validation (trailing slash on an empty path, lowercased host, dropped default port, punycode, percent-encoding), so the two clients return different strings for the same field.

See apify/apify-client-python#999 for the full context and apify/apify-client-python#1005 for how it was resolved there.

Decide whether the JS client should normalize as well. If yes, it belongs in 3.0 — it changes the values callers get back.

*✍️ Drafted by Claude Code*

vdusek:
Fields the OpenAPI specification marks as `format: uri` are validated with `z.url({ normalize: true })`, so the client returns the WHATWG URL serialization rather than the raw string the API sent. The Python client normalizes the same fields through `pydantic.AnyUrl`, so both clients now return the same string for the same field. On the Python side that normalization was kept and documented in apify/apify-client-python#1005, and this PR follows the same decision.

Normalization adds a trailing slash to an empty path, lowercases and punycodes the host, drops a default port, and percent-encodes unsafe characters. The specification carries `format: uri` on 17 properties, none of them on request-queue requests. Field types stay `string`, and a value that isn't a valid absolute URL now throws `ResponseValidationError`. `format: email` is left as a plain string, since the issue is about URL fields.

The v3 upgrading guide gets a "URL fields are normalized" section mirroring the Python one: the normalization cases, how to compare a field against a stored URL, and the affected field names. The error-handling page points at it.

Closes #1028

*✍️ Drafted by Claude Code*


barjin (approved):
lgtm 👍  if we're already doing this in Python, I don't really see an issue here. 

Thanks @vdusek !

PR #1046, opened 2026-09-08
feat!: expose the timeout tiers on the public API

### LINKED ISSUE #1029: Expose the timeout tiers on the public API
The client already picks a timeout tier per method internally (`SMALL` / `MEDIUM` / `DEFAULT` in `src/base/resource_client.ts`), but the only public knob is the single global `timeoutSecs`. There is no per-call override for a request that legitimately takes longer, and no way to change a tier's default.

Python: apify/apify-client-python#653 introduced the tiers with a per-call `timeout` argument and constructor overrides, apify/apify-client-python#664 tuned the defaults, apify/apify-client-python#962 warns when a requested timeout is capped.

*✍️ Drafted by Claude Code*

vdusek:
### Issue

- Closes #1029
- This is a port of the following:
  - apify/apify-client-python#653
  - apify/apify-client-python#664
  - apify/apify-client-python#962

### Description

The client already chose a timeout per method internally, but the only public knob was the global `timeoutSecs`, so a call that legitimately takes longer had no way to ask for more time. The tiers are now public, following the Python client, and every method that sends a request takes a `timeoutSecs` option.

| Tier | Default | Covers |
| --- | --- | --- |
| `short` | 5 s | metadata reads and writes |
| `medium` | 30 s | listing, batch and trigger calls |
| `long` | 360 s | downloads, uploads and streaming |
| `noTimeout` | none | polling that waits for a job to finish |

Every method keeps the tier its Python counterpart has. `timeoutMaxSecs` caps a single attempt and bounds the growth across retries, which doubles from the requested value (T, 2T, 4T, 8T). The six `waitForFinish` methods get the API's one-minute hold on top of their tier, so the client doesn't abort a request the API is still holding and start a second run on retry.

`timeoutSecs` takes a tier name, a number of seconds, or `'noTimeout'`. The unit lives in the option name rather than in the value, the way `waitSecs` and `timeoutMaxSecs` already do.

The last commit also drops the leading underscore from non-public members, as agreed in review. Where the underscore was covering a collision with a public member, the helper got a new name: `_url` to `buildUrl`, `_params` to `buildParams`, `_get` to `getResource`, `_list` to `listResources`, `_waitForFinish` to `waitForJobFinish`, `_batchAddRequests` to `addRequestBatch`, and `ApifyClient._options` to `subClientOptions`. `src/apify_api_error.ts` and `src/statistics.ts` are left for #1057.

## Testing

`test/client_timeouts.test.ts` covers the tier defaults, per-call overrides, the `waitForFinish` hold, `runTimeoutSecs` and the `requestQueue(id, { timeoutSecs })` cap. `test/http_client.test.ts` covers the cap warning and the retry growth. `test/apify_api_error.test.ts` guards `clientMethod`, which reads the renamed helpers off the stack.

## Breaking changes

- `timeoutSecs` on the constructor is replaced by `timeoutShortSecs`, `timeoutMediumSecs`, `timeoutLongSecs` and `timeoutMaxSecs`.
- Outside the storage clients, methods used the global 360 s and now use `short` or `medium`, so a call that waited out a slow API can fail sooner.
- The run timeout of `ActorClient.start()` / `call()`, `TaskClient.start()` / `call()` and `RunClient.resurrect()` is now `runTimeoutSecs`. TypeScript reports a leftover `timeout` at compile time, and the option schemas are strict, so JavaScript gets an `ArgumentValidationError` instead of a silently dropped run limit.
- `client.httpClient.call()` takes `timeoutSecs` in seconds, where the axios field it replaces took milliseconds. A direct call that passed `timeout: 30000` becomes `timeoutSecs: 30`.
- Methods without options gained an options parameter, and methods that take a payload gained a second one.
- The `protected` helpers of `ApiClient`, `ResourceClient` and `ResourceCollectionClient` are renamed, which reaches anyone subclassing them.

`KeyValueStoreClient.setRecord()` keeps its `timeoutSecs`, and a plain number still means what it did in v2, so Crawlee's `setValue()` needs no paired PR. The option now takes a tier name or `'noTimeout'` as well. The same goes for `client.requestQueue(id, { timeoutSecs })`, which caps the tier of every request that queue client sends.

*✍️ Drafted by Claude Code*

### DIFF
--- a/docs/public-api/apify-client.api.md
+++ b/docs/public-api/apify-client.api.md
@@ -363,7 +363,7 @@ export interface ActorStartOptions extends TimeoutOptions {
     maxTotalChargeUsd?: number;
     memory?: number;
     restartOnError?: boolean;
-    runTimeout?: number;
+    runTimeoutSecs?: number;
     waitForFinish?: number;
     webhooks?: readonly WebhookUpdateData[];
 }
@@ -633,7 +633,7 @@ interface ApifyRequestConfig extends Omit<AxiosRequestConfig, 'timeout'> {
     forceBuffer?: boolean;
     // (undocumented)
     stringifyFunctions?: boolean;
-    timeout?: Timeout;
+    timeoutSecs?: Timeout;
 }
 
 // Not exported by the entry point; reachable only as a referenced type.
@@ -3520,11 +3520,11 @@ export interface RequestQueueUserOptions {
 // Not exported by the entry point; reachable only as a referenced type.
 // @public
 class ResourceClient extends ApiClient {
-    protected deleteResource(timeout: Timeout): Promise<void>;
-    protected getResource<T, R>(schema: z.ZodType, options: T, timeout: Timeout): Promise<R | undefined>;
-    protected timeoutForWaitForFinish(timeout: Timeout | undefined, tier: TimeoutTier, waitForFinishSecs: number | undefined): Timeout;
+    protected deleteResource(timeoutSecs: Timeout): Promise<void>;
+    protected getResource<T, R>(schema: z.ZodType, options: T, timeoutSecs: Timeout): Promise<R | undefined>;
+    protected timeoutForWaitForFinish(timeoutSecs: Timeout | undefined, tier: TimeoutTier, waitForFinishSecs: number | undefined): Timeout;
     // (undocumented)
-    protected updateResource<T, R>(schema: z.ZodType, newFields: T, timeout: Timeout): Promise<R>;
+    protected updateResource<T, R>(schema: z.ZodType, newFields: T, timeoutSecs: Timeout): Promise<R>;
     protected waitForJobFinish<R extends {
         status: (typeof ACT_JOB_STATUSES)[keyof typeof ACT_JOB_STATUSES];
     }>(schema: z.ZodType, options?: WaitForFinishOptions): Promise<R>;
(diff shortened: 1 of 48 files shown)


src/base/resource_client.ts
```diff
@@ -14,29 +15,44 @@ import { ApiClient } from './api_client.js';
...
     /**
      * A 404 resolves to `undefined` only when the client names its resource by ID. A chained client without one, such
      * as `run.dataset()`, throws it instead (see `catchNotFoundForResourceOrThrow()`).
      */
-    protected async _get<T, R>(
-        schema: z.ZodType,
-        options: T = {} as T,
-        timeoutMillis?: number,
-    ): Promise<R | undefined> {
+    protected async _get<T, R>(schema: z.ZodType, options: T, timeout: Timeout): Promise<R | undefined> {
```
janbuchar: Side note — in crawlee v4, we decided to get rid of underscore prefixes and suffixes on non-public members. Perhaps we should do it here as well.

docs/public-api/apify-client.api.md
```diff
@@ -3917,6 +3925,17 @@ export interface TieredPricingPerEvent {
 export interface TieredPricingPerEventEntry extends GeneratedTieredPricingPerEventEntry {
 }
 
+// @public
+export type Timeout = TimeoutTier | 'noTimeout' | number;
```
janbuchar, line 3929: Accepting a number here loses the information about the unit (seconds/milliseconds/weeks/...)

src/base/resource_client.ts
vdusek: Okay, the code we touched here has been updated in this way. And the rest will be handled as part of https://github.com/apify/apify-client-js/issues/1057.

vdusek: Done, in the files this PR already touches: the base clients, `HttpClient`, `ApifyClient` and the resource clients. Where the underscore was covering a collision with a public member, the helper got a new name: `_url` to `buildUrl`, `_params` to `buildParams`, `_get` to `getResource`, `_list` to `listResources`, `_waitForFinish` to `waitForJobFinish`, `_batchAddRequests` to `addRequestBatch`, and `ApifyClient._options` to `subClientOptions`.

One thing the rename shook out: `ApifyApiError.clientMethod` scraped the stack with `\._?([A-Za-z]+)`, so the underscore was exactly what turned the internal frame `ActorCollectionClient._list` into the reported `list`. A public method that does `return this.getResource(...)` without awaiting leaves only the helper's frame on the async stack, so errors started reporting `updateResource`. There is a `PUBLIC_METHOD_BY_HELPER` map in `apify_api_error.ts` now, and `test/apify_api_error.test.ts` catches it.

`src/apify_api_error.ts` and `src/statistics.ts` are the two files left over, tracked in #1057 so this PR does not collide with the other branches based on `v3`. `LoggerActorRedirect._log()` stays as it is, since it overrides `Logger` from `@apify/log`.

*✍️ Drafted by Claude Code*

docs/public-api/apify-client.api.md
vdusek: Fair. The option is `timeoutSecs` now, so the unit sits in the name, the way `waitSecs` and `timeoutMaxSecs` already do. `Timeout` itself is unchanged, `TimeoutTier | 'noTimeout' | number`, with the number read as seconds.

`runTimeout` became `runTimeoutSecs` for the same reason, and that has a welcome side effect: `KeyValueStoreClient.setRecord()` keeps its v2 `timeoutSecs` and a plain number still means what it meant, so Crawlee's `setValue()` needs no paired PR after all.

*✍️ Drafted by Claude Code*

janbuchar (approved):

PR #1049, opened 2026-09-09
fix!: correct Actor version, enum and setRecord input types

### LINKED ISSUE #526: Wrong and missing types on `ActorVersion`, builds and `setRecord`
While rewriting the tests for apify-cli, I've encountered several issues with the types for ActorVersion (and I am partly to blame):

- enums are not exported but are used in concrete types (this is very bad as we don't provide string fallbacks, and thus people require casts)
  - solutions: either export the enum and use it, reference the types as imported from `@apify/consts`, or move to raw strings only
- ActorVersionSourceFile seems to be missing the `folder` boolean (this type is used in ActorVersionSourceFiles, and seems to signal a folder of files)
- keyValueStore(store.id).setRecord doesn't allow buffers
- version(version).update has a wrong type (also reported https://apify.slack.com/archives/C0L33UM7Z/p1709550430797429)
- builds lack newer fields, like `buildNumber`, `build.stats?.computeUnits` and `build.stats?.imageSizeBytes` being concretely typed

vdusek:
### Description

Three of the five bullets in #526 were still open on `v3`. `ActorVersionSourceFile.folder` and the newer build fields (`buildNumber`, `stats.computeUnits`, `stats.imageSizeBytes`) were already fixed there; these three were not.

- Inputs that took a published enum now take that enum's values as plain string literals, so `sourceType: 'GIT_REPO'` compiles without a cast. Four positions are affected: `ActorVersion.sourceType`, a scheduled action's `type`, `ActorCollectionListOptions.sortBy`, and the format `downloadItems()` takes. All four enums stay published and their members stay assignable, and the version union still rejects a source location that doesn't match the declared type.
- `setRecord()` takes `KeyValueStoreRecordValue`, which covers the buffers, typed arrays and streams the runtime has always uploaded. The existing buffer tests carried `as any` for exactly that reason, and those casts are gone. `Buffer` isn't spelled out in the union, because it's a `Uint8Array` and `TypedArray` already covers it.
- `ActorVersionClient.update()` takes `ActorVersionUpdateData`, a partial version. The spec posts `CreateOrUpdateVersionRequest` with every field optional, and the endpoint documents that it leaves untouched whatever the payload omits, so `update({ buildTag: 'latest' })`, the method's own JSDoc example, now compiles. `create()` keeps the full union, since the API requires `versionNumber` and `sourceType` there.

### Issue

Closes #526

### Breaking changes

- A variable annotated as `ActorSourceType`, `ScheduleActions`, `ActorListSortBy` or `DownloadItemsFormat` no longer accepts the matching value read off a client. Annotate it as `ActorVersion['sourceType']`, and so on, or leave it to inference. `setRecord()` also rejects an unsupported value with a new message, `Expected a JSON-serializable value, binary data, or a stream`.
- The v3 upgrading guide and the public API report are updated to match.

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -250,6 +250,10 @@ type GeneratedEnvVar = Schemas['EnvVar'];
  * Declared here rather than in `./resource_clients/actor_version` so that the version types can
  * reference it without closing an import cycle. It is re-exported from there, so the public name and
  * import path are unchanged.
+ *
+ * The fields that carry a source type are typed as `` `${ActorSourceType}` ``, the enum's values as
+ * plain string literals, rather than as the enum itself. A member stays assignable to that, so
+ * `ActorSourceType.GitRepo` keeps working, while `'GIT_REPO'` no longer needs a cast.
  */
 export enum ActorSourceType {
     SourceFiles = 'SOURCE_FILES',
@@ -288,7 +292,7 @@ export interface ActorVersionClientNarrowings {
     // The spec permits `sourceType: null`. It is deliberately not adopted: the published type is a
     // union discriminated on exactly this field, and a version with no source type carries no usable
     // source location either, so accepting the `null` would only make every variant unreachable.
-    sourceType: ActorSourceType;
+    sourceType: `${ActorSourceType}`;
 }
 
 /**
@@ -298,7 +302,7 @@ export interface ActorVersionClientNarrowings {
  * keeps a union discriminated on `sourceType` instead, because that narrows the source location down
  * to the single field which applies -- so the four are dropped here and reinstated per variant.
  */
-export interface BaseActorVersion<SourceType extends ActorSourceType>
+export interface BaseActorVersion<SourceType extends `${ActorSourceType}`>
     extends
         Omit<
             GeneratedVersion,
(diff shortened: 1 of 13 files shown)


src/models.ts
```diff
@@ -250,6 +250,10 @@ type GeneratedEnvVar = Schemas['EnvVar'];
  * Declared here rather than in `./resource_clients/actor_version` so that the version types can
  * reference it without closing an import cycle. It is re-exported from there, so the public name and
  * import path are unchanged.
+ *
+ * The fields that carry a source type are typed as `` `${ActorSourceType}` ``, the enum's values as
```
janbuchar: No need to teach typescript in the comments 🙂

test/actors.test.ts
```diff
@@ -55,8 +55,11 @@ describe('Actor methods', () => {
                 offset: 3,
                 desc: true,
                 my: true,
-                sortBy: ActorListSortBy.CREATED_AT,
+                sortBy: 'createdAt' as const,
```
janbuchar: Maybe `opts` could be inlined so that this isn't necessary for the code to typecheck?

janbuchar (approved):
LGTM

src/models.ts
vdusek: Yes, removed, and Claude should remember that

PR #1050, opened 2026-09-09
fix!: bump proxy-agent to v8 to remove the url.parse() deprecation warning

### LINKED ISSUE #898: `url.parse()` deprecation warning on Node 24 from proxy-agent 6
Full warning message:
```
[DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
```
I believe the warming comes via the proxy-agent@6.5 which depends on proxy-from-env@1.1, which addressed the warning in their 2.0 release.  Based on https://github.com/TooTallNate/proxy-agents/pull/400/changes I think upgrading to proxy-agent 7 or 8 should resolve the warning.

vdusek:
`proxy-agent@6` nests `proxy-from-env@1`, which runs every request URL through the legacy `url.parse()`. Node 24 turned that into runtime deprecation `DEP0169`, printed once per process. `proxy-agent@8` depends on `proxy-from-env@2`, which uses the WHATWG URL API. The major is ESM-only and needs Node 20+. `v3` is already native ESM on Node 22+, so no source changes.

`proxy-from-env@2` no longer reads `npm_config_proxy` and `npm_config_https_proxy`, the variables npm exports into `npm run` scripts from `.npmrc`, so a proxy configured only there stops applying to the client's requests. The v3 upgrading guide covers it.

Two new tests cover the proxy path, which nothing exercised before: the forward proxy named in `HTTP_PROXY` with a `NO_PROXY` exemption, and the CONNECT tunnel an https origin takes.

BREAKING CHANGE: proxy settings that reach the client through npm config are no longer honored. `proxy-from-env@2` reads `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY` and `NO_PROXY` and no longer consults `npm_config_proxy` or `npm_config_<proto>_proxy`, so a proxy configured only in `.npmrc` stops applying to requests made from an npm script. Set the standard environment variables instead.

Closes #898

*✍️ Drafted by Claude Code*


barjin (approved):
lgtm, thanks @vdusek !

PR #1051, opened 2026-09-09
perf!: serialize requests once in batchAddRequests

### LINKED ISSUE #972: Avoid double JSON serialization in `batchAddRequests`
Follow-up from the Python client's equivalent fix — see [apify/apify-client-python#953 and the discussion there](https://github.com/apify/apify-client-python/pull/953#discussion_r3601433178). The JS client has the same inefficiency.

In `batchAddRequests`, every request is JSON-serialized twice:

1. Once to measure payload size when splitting into batches — `sliceArrayByByteLength` (`src/utils.ts`) calls `JSON.stringify` on the whole slice, and on the splitting path on each request individually.
2. Once more when sending — the batch is passed as `data` to `httpClient.call`, and the `serializeRequest` axios interceptor (`src/interceptors.ts`) stringifies the same objects again to build the request body.

That's redundant CPU work in a hot path — Crawlee's `RequestQueue` goes through this method.

Suggested fix, mirroring apify/apify-client-python#953: serialize each request once up front, split batches by the serialized byte lengths (counting the `,` separators and the `[]` brackets), and send the pre-assembled JSON string as the body with an explicit `content-type: application/json` header so the interceptor doesn't re-serialize it. As a side effect this also fixes the current measurement slightly undercounting the actual body: it accounts for the `[]` brackets but not the commas between items (harmless today thanks to the safety buffer, but fixed for free by this approach).

*✍️ Drafted by Claude Code*

vdusek:
### Description

- Mirroring https://github.com/apify/apify-client-python/pull/953.
- `batchAddRequests` stringified every request twice: once in `sliceArrayByByteLength` to measure the batch, and again in the `serializeRequest` interceptor when sending. Axios then ran its default `transformRequest` on top, because `transformRequest: undefined` in the instance config falls back to the defaults, and that default validates a JSON string body by parsing it in full. Each batch body was stringified twice and parsed once more.
- Each request is now serialized once up front. The byte lengths decide the batch boundaries (commas and brackets counted, which the old measurement skipped) and the same strings are joined into the batch body, sent with an explicit `content-type: application/json`. The interceptor passes a string body with an explicit content type through untouched, and the axios instance sets `transformRequest` and `transformResponse` to `[]`. That also removes the validation re-parse for every JSON body the client sends.

### Issue

- Closes #972

### Breaking changes

- A string body declared as JSON but not valid JSON (for example `setRecord` with `contentType: 'application/json'` and a non-JSON string) is sent as it is. Axios used to double-encode it into a JSON string literal.
- A request too large for the payload limit is rejected before any batch is sent. It used to be detected only when its batch came up, with earlier batches already in flight.
- The protected `_batchAddRequests` and `_batchAddRequestsWithRetries` take the serialized entries, and `_batchAddRequests` no longer re-validates a batch the public method already validated. The API report is updated.

*✍️ Drafted by Claude Code*

### DIFF
--- a/docs/04_upgrading/upgrading_v3.md
+++ b/docs/04_upgrading/upgrading_v3.md
@@ -216,13 +216,9 @@ A trailing slash appears only on a field the API returns without a path, so on `
 
 A URL field whose value isn't a valid absolute URL now fails response validation and throws <ApiLink to="class/ResponseValidationError">`ResponseValidationError`</ApiLink>, the same as any other field that doesn't match the specification.
 
-## `batchAddRequests()` serializes each request once
+## `batchAddRequests()` rejects an oversized request before sending anything
 
-<ApiLink to="class/RequestQueueClient#batchAddRequests">`batchAddRequests()`</ApiLink> stringified every request twice, once to measure which batch it belonged to and once to send it. Each request is now serialized once, and the same strings are joined into the batch body. The bytes on the wire are the same. Three things around them changed.
-
-### An oversized request is rejected before any batch is sent
-
-A request too large for the payload limit still throws, and the message still names its index. What changed is when the check runs. In v2 each batch was measured as it came up, so every batch before the oversized request had already been sent by the time the call threw. v3 measures the whole input up front, so the call throws without sending anything.
+<ApiLink to="class/RequestQueueClient#batchAddRequests">`batchAddRequests()`</ApiLink> now measures the whole input before it sends the first batch. A request too large for the payload limit still throws the same error, and the message still names the request's index. What changed is that nothing has been sent by the time it throws. In v2 each batch was measured as it came up, so every batch before the oversized request had already gone out.
 
 ```js
 // In v2 the four batches before the oversized request had already been sent.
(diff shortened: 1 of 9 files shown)


src/interceptors.ts
```diff
@@ -43,6 +43,12 @@ function getHeader(config: ApifyRequestConfig, name: string): string | undefined
 }
 
 function serializeRequest(config: ApifyRequestConfig): ApifyRequestConfig {
+    // A string body with an explicit content type is already serialized and goes out as it is. The axios default
+    // transform would otherwise parse a JSON one in full just to check that it is valid, which for a body assembled
+    // from thousands of pre-serialized requests costs about as much as serializing them did.
+    const explicitContentType = config.headers?.['Content-Type'] || config.headers?.['content-type'];
```
barjin: `getHeader(config, 'content-type')` might be easier to read

docs/04_upgrading/upgrading_v3.md
```diff
@@ -272,6 +272,37 @@ A trailing slash appears only on a field the API returns without a path, so on `
...
+await client.requestQueue('my-queue').batchAddRequests([...hundredRequests, oversizedRequest]);
+```
+
+Code that treated the throw as "some of these landed" and reconciled the queue afterwards can drop the reconciliation.
+
+## A string body with a JSON content type is sent as it is
+
+A string body sent with an explicit `application/json` content type now goes out verbatim. Axios used to parse it to check that it was valid JSON, and wrapped it in a JSON string literal when it wasn't. The client skips that step so that `batchAddRequests()` can send a body it has already serialized.
+
+<ApiLink to="class/KeyValueStoreClient#setRecord">`setRecord()`</ApiLink> is where you'd notice. Storing a non-JSON string under `contentType: 'application/json'` used to save `"my value"`, quotes included, and now saves `my value`, which <ApiLink to="class/KeyValueStoreClient#getRecord">`getRecord()`</ApiLink> can't parse back:
```
barjin, line 291: > which <ApiLink to="class/KeyValueStoreClient#getRecord">`getRecord()`</ApiLink> can't parse back

This is kinda annoying, but understandable. 

I'm not sure whether we should automatically parse the KVS contents based on the `content-type` in the client, but that ship has sailed, I suppose.

barjin (approved):
Just some nits / thinking out loud, otherwise lgtm.

Thanks @vdusek !

docs/04_upgrading/upgrading_v3.md
vdusek: Yeah, I'm merging this. However, feel free to start a follow-up Slack discussion about it. I'd probably lean toward dropping this automatic parsing of KVS records based on their content type, but I think it deserves a bit more investigation and discussion first with a broader audience (also `#product-dev-tools` channel). I opened a follow-up issue for it: https://github.com/apify/apify-client-js/issues/1067, and we should discuss it before the v3 release.

PR #1052, opened 2026-09-09
feat!: make the client's own code runtime-independent

### LINKED ISSUE #753: Make `apify-client` runtime-independent

vdusek:
## Description

The client's own code imports no Node.js built-in. The Node-only parts, keep-alive HTTP agents with proxy support and request body compression, live in `src/runtime/node.ts`, with a Web API twin in `src/runtime/web.ts`. The `#runtime` entry in `imports` picks between them: the `node` condition resolves to the Node.js module, everything else to the Web API one, so a bundler targeting a browser or an edge runtime never sees `node:zlib`, `node:os`, `node:util` or `proxy-agent`. Node.js, Bun and Deno keep today's behavior.

Base64 encoding, byte counting, log chunk joining and response decoding go through `TextEncoder`, `TextDecoder` and `btoa`, so the `isNode()` sniff and the `BROWSER_BUILD` and `VERSION` globals are gone. rsbuild also needs `aliasStrategy: 'prefer-alias'`, so the `tsconfig.json` `paths` entry stops beating the `#runtime` alias. The bundle is 324 kB, down from 335 kB.

## Issues

- Closes #753. 
- `@apify/log` and `@apify/utilities` still import `node:events`, `node:stream` and `node:crypto`, so bundling the ES module build for a non-Node.js target needs polyfills for `events`, `process`, `stream` and `buffer`; tracked in apify/apify-shared-js#537.

## Testing

- `test/runtime.test.ts` covers both implementations and bundles `dist` with esbuild to assert that browser, Cloudflare Workers, neutral and Node.js targets each resolve `#runtime` to the right file, and that the Web API build imports no `node:` module.

## Breaking changes

- `TextDecoder` decodes response bodies instead of `Buffer`, so a `content-type` charset can be handled differently: `iso-8859-1` and similar decode to a string, `hex` and `base64` come back as raw bytes, `ascii` is read as `windows-1252`, and a leading UTF-8 BOM is stripped.
- Request compression now accepts `Uint8Array` and `ArrayBuffer` bodies, which gain a `content-encoding` header.
- Anything asking for the `browser` condition gets the Web API implementation even on Node.js. Jest's `jsdom` environment does, so a suite running under it loses log streaming, the `stream` record option, proxy support and compression unless `testEnvironmentOptions.customExportConditions` is set to `['node']`.

The v3 upgrading guide covers all of these.

*✍️ Drafted by Claude Code*

### DIFF
--- a/docs/04_upgrading/upgrading_v3.md
+++ b/docs/04_upgrading/upgrading_v3.md
@@ -178,9 +178,11 @@ Two return types change as a result of describing what the endpoints really retu
 
 ## The client's own code no longer needs Node.js
 
-The client's own code runs on Web APIs. The parts that need Node.js built-ins, the keep-alive HTTP agents with proxy support and request body compression, live in a module that the `#runtime` entry of the package's `imports` field selects at bundle time. The `node` condition gets the Node.js implementation, and every other target gets the Web API one, so a bundler targeting a browser or an edge runtime no longer pulls `node:zlib`, `node:os`, `node:net`, or `proxy-agent` out of the client. Two dependencies still import Node.js built-ins, so bundling the ES module build for a non-Node.js target still needs a few polyfills. For details, see [Bundled environments](../02_concepts/05_bundled-environments.md).
+The client's own code runs on Web APIs. The parts that need Node.js built-ins, the keep-alive HTTP agents with proxy support and request body compression, live in a module that the `#runtime` entry of the package's `imports` field selects at bundle time. The `node` condition gets the Node.js implementation, and every other target gets the Web API one, so a bundler targeting a browser or an edge runtime no longer pulls `node:zlib`, `node:os`, `node:util`, or `proxy-agent` out of the client. Two dependencies still import Node.js built-ins, so bundling the ES module build for a non-Node.js target still needs a few polyfills. For details, see [Bundled environments](../02_concepts/05_bundled-environments.md).
 
 On Node.js nothing changes. The features that need it, log streaming and the `stream` record option, proxy support and request compression, work as before.
+
+The pick happens when the import is resolved rather than at runtime, so a Node.js application bundled for a browser or a neutral target gets the Web API implementation and loses those features. The client used to sniff the runtime and recover. The bundler's target now decides.
 
 ### Response bodies are decoded by `TextDecoder`
 
--- a/docs/04_upgrading/upgrading_v3.md
+++ b/docs/04_upgrading/upgrading_v3.md
@@ -317,27 +317,37 @@ export HTTPS_PROXY=http://proxy.example.com:3128
 
 The same `proxy-agent` upgrade removes the `[DEP0169] DeprecationWarning` about `url.parse()` that Node.js 24 and newer printed on the client's first request.
 
-## The client's own code no longer needs Node.js
+## Node.js-only features follow the bundler's target
 
-The client's own code runs on Web APIs. The parts that need Node.js built-ins, the keep-alive HTTP agents with proxy support and request body compression, live in a module that the `#runtime` entry of the package's `imports` field selects at bundle time. The `node` condition gets the Node.js implementation, and every other target gets the Web API one, so a bundler targeting a browser or an edge runtime no longer pulls `node:zlib`, `node:os`, `node:util`, or `proxy-agent` out of the client. Two dependencies still import Node.js built-ins, so bundling the ES module build for a non-Node.js target still needs a few polyfills. For details, see [Bundled environments](../02_concepts/05_bundled-environments.md).
+Four features need Node.js APIs: log streaming with <ApiLink to="class/LogClient#stream">`LogClient.stream()`</ApiLink> and <ApiLink to="class/RunClient#getStreamedLog">`RunClient.getStreamedLog()`</ApiLink>, the `stream` option of <ApiLink to="class/KeyValueStoreClient#getRecord">`KeyValueStoreClient.getRecord()`</ApiLink>, proxy support, and request body compression.
 
-On Node.js nothing changes. The features that need it, log streaming and the `stream` record option, proxy support and request compression, work as before.
+In v2 the client detected the runtime when one of them was used, so a Node.js application bundled for a browser or a neutral target kept them all. In v3 the implementation is picked when the import is resolved, so the bundler's conditions decide. Without the `node` condition, `getRecord({ stream: true })` throws, `getStreamedLog()` returns `undefined`, and requests go out unproxied, uncompressed, and without the client's `User-Agent` header.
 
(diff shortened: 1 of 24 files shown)


src/body_parser.ts
```diff
@@ -1,67 +1,56 @@
...
     return contentType === CONTENT_TYPE_JSON ? JSON.parse(dataString) : dataString;
 }
 
-export function isomorphicBufferToString(buffer: Buffer | ArrayBuffer, encoding: BufferEncoding): string {
-    if (buffer.constructor.name !== ArrayBuffer.name) {
-        return buffer.toString(encoding);
+function createDecoder(charset?: string): TextDecoder | undefined {
+    try {
+        // No charset: hope that it's utf-8.
+        return new TextDecoder(charset || 'utf-8');
```
B4nan: `TextDecoder` defaults to `ignoreBOM: false`, so every decoded body loses a leading U+FEFF. A BOM-prefixed CSV stored in a KVS (common when the file came from Excel) then round-trips through `getRecord()` differently than in v2. `ignoreBOM: true` keeps the v2 bytes and drops a paragraph from the upgrading guide. The `|| 'utf-8'` fallback is redundant too, since `new TextDecoder(undefined).encoding` is already `'utf-8'`.

```suggestion
        // No charset defaults to utf-8. Keep a leading BOM, as v2's `Buffer` decoding did.
        return new TextDecoder(charset, { ignoreBOM: true });
```

test/body_parser.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        expect(maybeParseBody(body, 'text/plain; charset=ascii')).toBe('café');
+    });
+
+    test('strips a leading byte order mark from a JSON body, which would otherwise fail to parse', () => {
+        expect(maybeParseBody(encode('﻿{"a":1}'), 'application/json')).toEqual({ a: 1 });
+    });
+
+    test('strips a leading byte order mark from a text body', () => {
+        expect(maybeParseBody(encode('﻿a,b\n1,2'), 'text/csv; charset=utf-8')).toBe('a,b\n1,2');
+    });
```
B4nan: With `ignoreBOM: true` these two become one test for the v2 behavior.

```suggestion
    test('keeps a leading byte order mark in a text body, as v2 did', () => {
        expect(maybeParseBody(encode('﻿a,b\n1,2'), 'text/csv; charset=utf-8')).toBe('﻿a,b\n1,2');
    });
```

docs/04_upgrading/upgrading_v3.md
```diff
@@ -404,4 +404,39 @@ The `clientMethod` field on <ApiLink to="class/ApifyApiError">`ApifyApiError`</A
...
+```
+
+Running on Node.js without a bundler is unaffected, and the pre-built browser bundle behaves as it did in v2. For what bundling for a non-Node.js target takes, see [Bundled environments](../02_concepts/05_bundled-environments.md).
+
+### Response bodies are decoded by `TextDecoder`
+
+In Node.js, v2 decoded response bodies with `Buffer` and v3 decodes them with `TextDecoder`. The two support different charsets, so a `content-type` header carrying one can be handled differently:
+
+- `iso-8859-1` and other charsets only `TextDecoder` knows decode to a string, where v2 handed back raw bytes.
+- `hex` and `base64`, which only `Buffer` knows, come back as raw bytes, where v2 decoded the body as if it were in that encoding.
```
B4nan: `binary`, `ucs2` and `utf16le` are `Buffer`-only labels too (`new TextDecoder('utf16le')` throws, `utf-16le` is fine).

```suggestion
- `hex`, `base64`, `binary`, `ucs2` and `utf16le`, which only `Buffer` knows, come back as raw bytes, where v2 decoded the body as if it were in that encoding.
```

docs/04_upgrading/upgrading_v3.md
```diff
@@ -404,4 +404,39 @@ The `clientMethod` field on <ApiLink to="class/ApifyApiError">`ApifyApiError`</A
...
+- `ascii` is read as an alias for `windows-1252`, so a byte above `0x7F` decodes to the character that encoding gives it, where v2 masked the byte down to seven bits.
+- A leading UTF-8 byte order mark is stripped from every decoded body, so a JSON body carrying one parses instead of throwing, and a `text/*` record such as a BOM-prefixed CSV comes back without it.
```
B4nan: Drop this bullet once the decoder keeps the BOM.

```suggestion
```

docs/04_upgrading/upgrading_v3.md
```diff
@@ -404,4 +404,39 @@ The `clientMethod` field on <ApiLink to="class/ApifyApiError">`ApifyApiError`</A
...
+
+Code that depended on one of these has to convert the value itself. A body with no charset or with a UTF-8 one is unaffected, apart from the byte order mark, and that covers everything the Apify API sends.
```
B4nan:
```suggestion
Code that depended on one of these has to convert the value itself. A body with no charset or with a UTF-8 one is unaffected, and that covers everything the Apify API sends.
```

src/utils.ts
```diff
@@ -158,50 +164,45 @@ export function catchNotFoundForResourceOrThrow(err: ApifyApiError, resourceId:
...
-    data: Buffer;
-    encoding: 'br' | 'gzip';
+/**
+ * Views a request body as bytes: a string is UTF-8 encoded, binary values are viewed in place. Anything else
+ * - a stream, a `Blob`, form data - is `undefined`.
+ */
+function toBytes(value: unknown): Uint8Array | undefined {
+    if (typeof value === 'string') return textEncoder.encode(value);
+    if (ArrayBuffer.isView(value)) return new Uint8Array(value.buffer, value.byteOffset, value.byteLength);
+    if (value instanceof ArrayBuffer) return new Uint8Array(value);
```
B4nan: `isBuffer()` below uses tag checks on purpose, to also match an `ArrayBuffer` from another realm. This `instanceof` skips compression for such a body. Harmless, but using the same guard keeps the two in step.

```suggestion
    if (isBuffer(value)) return new Uint8Array(value);
```

B4nan (commented):
Looks good overall. A few suggestions below, mostly one-liners.

barjin (approved):
I suppose the changes in the decoding could cause some very niche edge cases (e.g., when reading KVS records), but it's overall a step in the right direction imo, so approving.

Thanks @vdusek !

PR #1055, opened 2026-09-10
fix: carry the API error body of a failed streaming request

### LINKED ISSUE #1043: A 404 from a streaming request produces an unusable ApifyApiError
`LogClient.stream()` and `KeyValueStoreClient.getRecord({ stream: true })` set `responseType: 'stream'`. When such a request fails, the `ApifyApiError` it produces carries nothing usable beyond `statusCode`. The message reads `Unexpected error: [object Object]`, and the `type`, `message` and `data` the API returned in the error body are all missing.

## Root cause

`parseResponseData` (`src/interceptors.ts:94`) bails for any response type other than `arraybuffer`, so a streamed error body is never read. `ApifyApiError` then receives an `IncomingMessage` as `response.data`, `JSON.stringify` throws on its circular references, and the constructor falls back to `Unexpected error: ${dataString}` (`src/apify_api_error.ts:100`).

One side effect: `catchNotFoundOrThrow` can never match a streamed 404, because it tests `err.type` against `record-not-found` and there is no type to test. `LogClient.stream()` therefore throws on a 404 even when the log is addressed by ID, and both the `undefined` arm of its return type and the `if (!logStream) return` guard in `StreamedLog` are unreachable.

## Expected behavior

An error from a streaming request should carry the same `type`, `message` and `data` as one from a regular request. Buffering the error body when `responseType` is `stream` would get there.

## Open questions

Parsing the body makes `catchNotFoundOrThrow` start matching, which flips `LogClient.stream()` back to resolving to `undefined` on a 404 for ID-addressed clients. #1042 documented and pinned the current throwing behavior, so either outcome is a deliberate choice that has to land across three places at once:

1. `src/resource_clients/log.ts`, both the `stream()` docstring and its `Promise<Readable | undefined>` return type.
2. `docs/04_upgrading/upgrading_v3.md`, the paragraph stating that `stream()` is not part of the v3 change.
3. `test/logs.test.ts`, the `stream() throws on 404 status code` case.

Found while reviewing #1042.

*✍️ Drafted by Claude Code*

vdusek:
Follow-up to #1041.

A failed request with `responseType: 'stream'` (`LogClient.stream()`, `getRecord({ stream: true })`) handed an unread `IncomingMessage` to `ApifyApiError`. `JSON.stringify` threw on its circular references, so the message read `Unexpected error: [object Object]` and the API's `type`, `message` and `data` were lost.

`HttpClient` now reads a streamed error body into a buffer before building the error, as the Python client does, using `concatStreamToBuffer` from `@apify/utilities`. A read that fails mid-body is swallowed: a raw socket error carries no status code, which would break `getRecord({ stream: true })` resolving to `undefined` on a 404. `ApifyApiError` decodes the buffer once and keeps a non-JSON body as text, so an HTML error page from a proxy is readable instead of a dump of byte values. That last part also improves `downloadItems()` and `getRecord({ buffer: true })` errors.

Known limitation, left as a follow-up: the body read sits outside the axios timeout, because axios settles a stream response at its headers. The agent's socket idle timeout bounds it, measured at 2043 ms under a 2 s client timeout, and `maxContentLength: -1` leaves it without a size cap. Tightening either needs a separate error-body budget.

Closes #1043

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_api_error.ts
+++ b/src/apify_api_error.ts
@@ -87,8 +87,8 @@ export class ApifyApiError extends Error {
         let responseData = response.data;
         let errorData: Record<string, unknown> | undefined;
 
-        // Some methods (e.g. downloadItems) set up forceBuffer on request response. If this request failed
-        // the body buffer needs to parse to get the correct error.
+        // A `forceBuffer` request (e.g. `downloadItems()`) and a failed streaming request, whose body `HttpClient`
+        // has read into a buffer, both arrive unparsed. Parse the body here to get at the error.
         if (isBuffer(responseData)) {
             try {
                 responseData = JSON.parse(isomorphicBufferToString(response.data, 'utf-8'));
--- a/src/apify_api_error.ts
+++ b/src/apify_api_error.ts
@@ -90,10 +90,12 @@ export class ApifyApiError extends Error {
         // A `forceBuffer` request (e.g. `downloadItems()`) and a failed streaming request, whose body `HttpClient`
         // has read into a buffer, both arrive unparsed. Parse the body here to get at the error.
         if (isBuffer(responseData)) {
+            const body = isomorphicBufferToString(response.data, 'utf-8');
             try {
-                responseData = JSON.parse(isomorphicBufferToString(response.data, 'utf-8'));
+                responseData = JSON.parse(body);
             } catch {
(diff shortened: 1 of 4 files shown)


barjin (approved):
lgtm, thanks 👍

PR #1058, opened 2026-09-10
feat!: allow passing a custom HTTP client to ApifyClient

### LINKED ISSUE #855: Allow passing a custom HTTP client to ApifyClient
Allow users to provide their own HTTP client implementation to `ApifyClient`, enabling customization for use cases not covered by the default client.

This was already implemented in the Python client:
- Issue: https://github.com/apify/apify-client-python/issues/416
- PR: https://github.com/apify/apify-client-python/pull/641

The Python implementation introduces abstract base classes that users can extend and pass via `ApifyClient(httpClient=...)`.

vdusek:
### Description

`ApifyClient` sends its requests through an HTTP client you can replace, following apify-client-python (apify/apify-client-python#641). The abstract `HttpClient` holds the shared pipeline: header merging, body serialization, compression, retries with backoff, timeout growth, body parsing and the `ApifyApiError` conversion. A transport implements `sendRequest()` and can override `isTimeoutError()`, `isRetryableTransportError()` and `close()`. `AxiosHttpClient` is the only built-in transport, and `async-retry` is no longer a dependency.

`ApifyClient.withCustomHttpClient({ token, baseUrl, publicBaseUrl, httpClient })` plugs a client in, as Python's `with_custom_http_client` does, since the retry and timeout options of the constructor configure the default client only. `ApifyClientOptions.headers` replaces the fixed-header use of `requestInterceptors`. `ApifyRequestConfig.signal` aborts a call, which then rejects with the signal's `reason`. `HttpClient` validates its retry and timeout options with the shape `ApifyClientOptions` uses.

Docs: an "HTTP clients" concept page, a "Build a custom HTTP client" guide over `fetch` with a section on reusing a Crawlee 4 client, the v3 upgrading guide and the regenerated api-extractor report. Tests drive a `call()`-only client and a hooks-only client over `node:http` through the pipeline.

### Issue

Closes #855

Follow-up: #1071

### Breaking changes

- `requestInterceptors` moved from `ApifyClientOptions` to `AxiosHttpClientOptions`. Use `headers` for a fixed header, or pass an `AxiosHttpClient` to `withCustomHttpClient()`. An interceptor now sees the body already serialized.
- `ApifyRequestConfig` and `ApifyResponse` no longer extend the axios types. `forceBuffer` became `responseType: 'buffer'`, and `responseType` takes `'parsed' | 'buffer' | 'stream'`.
- `InvalidResponseBodyError.response` is the transport's `HttpResponse` with the raw `body`.
- `ApifyApiError.httpMethod` is uppercase (`GET`).
- A `Blob`, `File`, `FormData` or web `ReadableStream` body throws a `TypeError`. Read it into a `Buffer` first.
- A form-encoded object body is encoded by the client: binary values are base64-encoded and nested arrays use `key[]`.
- The error thrown after the retries are exhausted is the last one, as in Python.
- The per-attempt timeout doubles on a fixed schedule up to `timeoutMaxSecs`, and retry delays are randomized.
- The `axios`, `httpAgent` and `httpsAgent` properties moved from `HttpClient` to `AxiosHttpClient`, and the instance no longer carries the default headers.

*✍️ Drafted by Claude Code*


src/http_clients/base.ts
```diff
@@ -0,0 +1,777 @@
...
+    /** HTTP method. */
+    method: HttpMethod;
+    /** Full request URL with the query string already encoded into it. */
+    url: string;
+    /** Final request headers, with the client's default headers already merged in. */
+    headers: Record<string, string>;
+    /** Request body, already serialized and compressed, or `undefined` for a request without one. */
+    body?: HttpRequestBody;
+    /** Timeout for this attempt in milliseconds. */
+    timeoutMillis: number;
```
janbuchar: Two things this shape closes off:

- No `signal` in `HttpRequest`/`ApifyRequestConfig`, so an in-flight call can only end via the per-attempt timeout. Crawlee v4's `BaseHttpClient` takes one in `SendRequestOptions` and composes it with the timeout.
- `HttpRequest`/`HttpResponse` are bespoke records rather than `Request`/`Response`, so no crawlee client (`FetchHttpClient`, `@crawlee/impit-client`) can be passed to `withCustomHttpClient()` as is. That is where proxying, TLS impersonation and fingerprinting already live; an SDK actor could push crawling and API traffic through one transport.

The adapter is ~40 lines: Node `Readable` → web stream plus `duplex: 'half'`, `getSetCookie()` to undo `Headers` folding repeated headers, and `isRetryableTransportError()` because crawlee's base classifies nothing. Auth headers, brotli bodies, streamed responses, 429/5xx retries, `ApifyApiError` and `noTimeout` all went through unchanged.

The gap is small but entirely in the type shape, so worth deciding now whether a fetch-shaped hook is the contract — `Request`/`Response` is also the more runtime-independent option after #1052.

Not arguing for crawlee's base as the abstraction: it owns cookies and redirects, which for API traffic means a fresh `tough-cookie` jar per request (~0.8 ms on loopback, for `Set-Cookie` nobody reads) and redirect following with no opt-out, losing the axios client's "no redirects for stream bodies" rule. Cheap version: `signal` plus the adapter as a documented recipe.

src/apify_client.ts
```diff
@@ -94,22 +103,32 @@ export class ApifyClient {
...
+    /**
+     * Statistics of the API calls made through the client. With a custom HTTP client, this is the client's own
+     * `stats` object, so it only reflects what that client records.
+     */
     stats: Statistics;
 
     logger: Log;
 
-    httpClient: HttpClient;
+    private _httpClient?: HttpClient;
```
janbuchar: `private _httpClient` / `private readonly _httpClientOptions` are the odd ones out — this class already has `#subClientOptions()` and `HttpClient`/`AxiosHttpClient` use `#` fields throughout.

Crawlee settled this in apify/crawlee#3108 with `no-underscore-dangle` (`enforceInClassFields`/`enforceInMethodNames`) plus `crawlee/prefer-private-fields`, a 33-line local oxlint plugin flagging the TS `private` modifier. The first is nearly free here: `oxlint --deny no-underscore-dangle src test scripts` reports only these two plus `_outputWithConsole`/`_log` in `LoggerActorRedirect`, which are inherited from `@apify/log` and belong on `allow`, as crawlee does for `_read` and friends.

src/apify_client.ts
```diff
@@ -94,22 +103,32 @@ export class ApifyClient {
...
 
+    /** Configuration of the default HTTP client, applied when no custom one is set. */
+    private readonly _httpClientOptions: HttpClientOptions;
+
+    /**
```
janbuchar, line 126: This is a footnote, not a description. No `@param`, nothing about what the constructor does (`ApifyClientOptions` validation, configuring the `AxiosHttpClient` that `httpClient` lazily creates), so typedoc renders "To use a custom HTTP client, use `withCustomHttpClient` instead" as the entire doc of the package's primary entry point.

Either document it properly or drop the block and put the pointer where someone choosing between the two constructors reads: the class doc or `ApifyClientOptions`.

docs/03_guides/02_custom-http-client.md
```diff
@@ -0,0 +1,68 @@
...
+
+const RETRYABLE_CODES = new Set(['ECONNREFUSED', 'ECONNRESET', 'EPIPE', 'ETIMEDOUT', 'EAI_AGAIN', 'UND_ERR_SOCKET']);
+
+class FetchHttpClient extends HttpClient {
+    async sendRequest({ method, url, headers, body, timeoutMillis, stream }) {
+        const response = await fetch(url, {
+            method,
+            headers,
+            body,
+            signal: AbortSignal.timeout(timeoutMillis),
```
janbuchar: `AbortSignal.timeout(0)` aborts on the next tick, and `computeTimeoutMillis()` returns `0` for `'noTimeout'` — the default of `ActorClient.call()`, `TaskClient.call()` and `ResourceClient._waitForFinish()`, hence `run.waitForFinish()` and `build.waitForFinish()`. Against a local server, `medium` goes through and `timeoutSecs: 'noTimeout'` dies with `TimeoutError: The operation was aborted due to timeout`; the `isRetryableTransportError()` override below retries it, so it fails nine times and looks like a network problem.

Same line in the `@example` on `HttpClient`.

`timeoutMillis ? AbortSignal.timeout(timeoutMillis) : undefined` fixes both, but the real fix is `HttpRequest.timeoutMillis`: it says "Timeout for this attempt in milliseconds", the concepts table says "Timeout of this attempt", and the `0` convention only appears in `computeTimeoutMillis()`'s JSDoc, which nobody writing a transport reads.

src/apify_api_error.ts
```diff
@@ -13,9 +13,14 @@ export type { ApifyApiErrorType } from './models.js';
...
  * 4: "listResources"
+ *
+ * The error is created inside the HTTP client's pipeline, whose frames sit above the resource client's. V8 names
+ * them after the transport class, which a custom client can call anything, so the second lookahead skips them by
+ * method name as well: no resource client has a method of either name.
  * @private
  */
-const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\.([A-Za-z]+) \(/;
+const CLIENT_METHOD_REGEX =
+    /at( async)? (?![A-Za-z]*HttpClient\.)([A-Za-z]+(Collection)?Client)\._?(?!makeRequest|retryWithExpBackoff)([A-Za-z]+) \(/;
```
janbuchar: Neither name in the skip list can match. V8 keeps the `#` in the frame and `\._?([A-Za-z]+)` has nothing to match it with:

```
at ApifyApiError.fromResponse (apify_api_error.js:123:16)
at #makeRequest (base.js:322:40)
at async #retryWithExpBackoff (base.js:253:24)
at async KeyValueStoreClient.deleteRecord (key_value_store.js:452:9)
```

(Node 26 against `dist`; under vitest the frames read `AxiosHttpClient.#makeRequest` — still `#`.) Cutting back to the v2 pattern with both lookaheads gone keeps `pluggable_http_client`, `apify_api_error` and `http_client` green — 81 tests, `PipelineClient` case included.

The frame that does leak is `call`, which the list doesn't name. Two clients with identical bodies, both `await super.call(config)`:

```
ImpitClient     -> clientMethod = ImpitClient.call
ImpitHttpClient -> clientMethod = KeyValueStoreClient.deleteRecord
```

Overriding `call()` is documented as supported; only the `HttpClient` suffix saves the second. Adding `call` fixes the attribution and gives the lookahead something to do.

src/http_clients/base.ts
```diff
@@ -0,0 +1,777 @@
...
+    readonly httpCompressor: HttpCompressor;
+
+    /**
+     * Headers sent with every request: the `Authorization` header built from the token, the `User-Agent` where the
+     * runtime lets a client set one, the `X-Apify-Workflow-Key` when a workflow key is configured, and whatever
+     * {@link HttpClientOptions.headers} added.
+     */
+    protected readonly defaultHeaders: Record<string, string>;
+
+    constructor(options: HttpClientOptions = {}) {
```
janbuchar, line 310: `ApifyClientOptions` validates these (`maxRetries` is `z.number().int().nonnegative()`); this constructor doesn't, and the concepts page and fetch guide both hand it options directly.

```
maxRetries = 8    attempt timeoutMillis = 30000
maxRetries = 0    attempt timeoutMillis = 30000
maxRetries = -1   attempt timeoutMillis = 15000
```

Negative skips the loop and runs the single attempt as `attempt = 0`, so `2 ** -1` halves the timeout, and a 429 there throws `ArgumentValidationError` out of `Statistics.addRateLimitError` (`gt(0)`). Fractional values give fractional exponents.

`minDelayBetweenRetriesMillis` is unguarded on both paths — `z.number()` without `.nonnegative()`, so a negative value retries with no delay. Pre-existing on `v3`, but it's the sibling check.

src/apify_client.ts
```diff
@@ -126,20 +145,75 @@ export class ApifyClient {
...
+     * {@link HttpClient.setDefaultAuthorization} and points {@link stats} at its statistics, so the counters keep
+     * tracking the calls the client actually makes. Resource clients hold on to the HTTP client they were created
+     * with, so assign before reaching for them.
+     */
+    get httpClient(): HttpClient {
+        this._httpClient ??= new AxiosHttpClient(this._httpClientOptions);
+        return this._httpClient;
+    }
+
+    set httpClient(httpClient: HttpClient) {
```
janbuchar, line 228: Reading `httpClient` before assigning materializes an `AxiosHttpClient`; if it sent anything, this drops it with keep-alive agents open and no handle left to `close()`. The getter doc warns about resource clients capturing the instance, not about this. Setter is sync, so a sentence in that doc is probably the whole fix.

src/http_clients/base.ts
```diff
@@ -0,0 +1,851 @@
...
+     * Actor input carrying page functions reaches the API.
+     */
+    stringifyFunctions?: boolean;
+    /** Give up on the first transport timeout instead of retrying it. */
+    doNotRetryTimeouts?: boolean;
+    /**
+     * Aborts the call. Once the signal aborts, the attempt in flight is ended, no retry follows, and the call
+     * rejects with the signal's `reason`.
+     */
+    signal?: AbortSignal;
```
janbuchar, line 140: The signal stops here. No resource-client method takes one — nothing in `src/base/` mentions `signal` — so a caller can abort `httpClient.call()` but not `client.actor('x').get()` or `run.waitForFinish()`, which is where someone with a shutdown handler would reach.

Fine as a follow-up.

src/apify_api_error.ts
```diff
@@ -13,9 +13,16 @@ export type { ApifyApiErrorType } from './models.js';
...
+ * The error is created inside the HTTP client's pipeline, whose frames sit above the resource client's. The
+ * pipeline's own methods are private, so V8 prints them with a `#` that no method name can match. A custom client
+ * can be named anything and can override the public `call()`, so its frames are skipped by the `HttpClient` suffix
+ * and by the method name. `ActorClient.call()` and `TaskClient.call()` reach the pipeline through `start()`, whose
+ * frame sits closer to the error and matches first.
  * @private
  */
-const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\.([A-Za-z]+) \(/;
+const CLIENT_METHOD_REGEX =
+    /at( async)? (?![A-Za-z]*HttpClient\.)([A-Za-z]+(Collection)?Client)\.(?!call \()([A-Za-z]+) \(/;
```
janbuchar, line 26: Works for the two shapes the tests cover, but the hole moved rather than closed: a transport whose class name ends in `Client` and whose `call()` awaits a helper of its own gets blamed on the helper. Running this regex against such a stack gives `ImpitClient.execute`, not `KeyValueStoreClient.deleteRecord`.

Nobody writes that today, and the fix as it stands is an improvement, so this is only worth noting because the comment above now claims the frames are skipped "by the `HttpClient` suffix and by the method name" — it is only `call` that is skipped.

src/http_clients/base.ts
```diff
@@ -0,0 +1,851 @@
...
+
+const CONTENT_TYPE_FORM_URLENCODED = 'application/x-www-form-urlencoded';
+
+/**
+ * Bounds of the numeric {@link HttpClientOptions}. `ApifyClientOptions` checks the same bounds and adds the defaults.
+ * @internal
+ */
+export const httpClientOptionsShape = {
+    maxRetries: z.number().int().nonnegative(),
+    minDelayBetweenRetriesMillis: z.number().nonnegative(),
```
janbuchar: `nonnegative()` here also tightens `ApifyClientOptions`, which took any number before. Nonsensical input either way, but it is a behaviour change in a PR whose breaking-changes list is otherwise exhaustive.

src/http_clients/base.ts
vdusek: follow-up issue - https://github.com/apify/apify-client-js/issues/1071

src/http_clients/base.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/2078f5881f27a8dac6cec65c66a0c492cc85fafc and https://github.com/apify/apify-client-js/commit/413defd9dea26a36cd7a333df7a3d329b1daa446

src/apify_client.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/b9fd7d068239832822e911658c956a265e43e720

src/apify_client.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/32310e3c14c6f95431ed2c12f80f6107598a9e12

docs/03_guides/02_custom-http-client.md
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/8ca5565293a563bef4713bd0430df4e9cb51ea3b

src/apify_api_error.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/22b7577b6d4aef59508a43661c1e3b09124f13f5

src/http_clients/base.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/4dfd4eab4eef6b33cd9a22ef0dc935f4eb07aad9

src/apify_client.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/32310e3c14c6f95431ed2c12f80f6107598a9e12

src/apify_api_error.ts
vdusek: Resolved in https://github.com/apify/apify-client-js/commit/679ab82c66ef6ded47766ea59300fda6fdd49ce4

docs/04_upgrading/upgrading_v3.md
```diff
@@ -486,3 +486,81 @@ Code that depended on one of these has to convert the value itself. A body with
...
+
+## The HTTP client is pluggable
+
+`ApifyClient` sends its requests through an HTTP client you can replace. <ApiLink to="class/HttpClient">`HttpClient`</ApiLink> is now the abstract base holding the shared request pipeline, and the axios-based client the `ApifyClient` uses by default is <ApiLink to="class/AxiosHttpClient">`AxiosHttpClient`</ApiLink>. A custom client extends `HttpClient`, implements `sendRequest()` and is plugged in with <ApiLink to="class/ApifyClient#withCustomHttpClient">`ApifyClient.withCustomHttpClient()`</ApiLink>. See [HTTP clients](../02_concepts/08_http-clients.md) for the contract. Decoupling the pipeline from axios changes a few details of the public surface.
+
+### `requestInterceptors` moved to `AxiosHttpClient`
+
+Axios request interceptors are a feature of the axios transport, so the option left `ApifyClientOptions`. Most interceptors only added a fixed header to every request, such as `x-apify-integration-platform`. The new `headers` option of `ApifyClientOptions` covers that without touching axios:
```
janbuchar: What do you mean, most interceptors?

janbuchar (approved):

PR #1059, opened 2026-09-10
refactor!: drop the underscore prefix from non-public members

### LINKED ISSUE #1057: Drop the underscore prefix from the remaining non-public members
Non-public members in this repo are marked with a leading underscore on top of the `private` or `protected` keyword. Crawlee v4 dropped that convention, and we agreed to follow it here ([review comment](https://github.com/apify/apify-client-js/pull/1046#discussion_r3978310776)).

#1046 renames the members in the files it already touches: the base clients, `HttpClient`, `ApifyClient` and the resource clients. Two files are left:

- `src/apify_api_error.ts` - `_safelyParsePathFromResponse()`, `_extractClientAndMethodFromStack()`, `_createApiStack()`
- `src/statistics.ts` - `_fillBlanksWithZeroes()`

`LoggerActorRedirect._log()` and `_outputWithConsole()` in `src/resource_clients/log.ts` stay as they are, since both come from the `Logger` base class in `@apify/log`.

Worth doing after #1046 lands, so the two changes don't collide.

*✍️ Drafted by Claude Code*

vdusek:
Non-public members in this repo carried a leading underscore on top of the `private` / `protected` keyword. Crawlee v4 dropped that convention and we agreed to follow it here.

This renames every such member across `src/`, so none are left. `LoggerActorRedirect._log()` and `_outputWithConsole()` in `src/resource_clients/log.ts` keep their names, since both come from the `Logger` base class in `@apify/log`. The code samples in `CONTRIBUTING.md` and the public API report follow the rename.

## Renamed helpers

The protected helpers on the base clients show up in the API report, so subclasses calling the old names have to switch. Where a bare name would collide with a public method, the helper got a suffix.

| Old | New |
| --- | --- |
| `_url` | `buildUrl` |
| `_publicUrl` | `buildPublicUrl` |
| `_params` | `buildParams` |
| `_subResourceOptions` | `subResourceOptions` |
| `_toSafeId` | `toSafeId` |
| `_listPaginatedFromCallback` | `listPaginatedFromCallback` |
| `_get` | `getResource` |
| `_update` | `updateResource` |
| `_delete` | `deleteResource` |
| `_waitForFinish` | `waitForJobFinish` |
| `_list` | `listResources` |
| `_listPaginated` | `listResourcesPaginated` |
| `_create` | `createResource` |
| `_getOrCreate` | `getOrCreateResource` |
| `_batchAddRequests` | `addRequestBatch` |
| `_batchAddRequestsWithRetries` | `addRequestBatchWithRetries` |

The names match what #1046 picked for the files it touches, so whichever of the two lands second rebases onto near-identical lines.

## ApifyApiError.clientMethod

`clientMethod` scrapes the stack with `\._?([A-Za-z]+)`, and the underscore was what turned an internal frame like `ActorCollectionClient._list` into the public-looking `list`. A public method that returns the promise of a shared helper without awaiting it leaves only the helper's frame on the async stack, so after the rename the field would report `listResources` to users. A `PUBLIC_METHOD_BY_HELPER` map in `apify_api_error.ts` maps helper frames back to the public method, and `test/apify_api_error.test.ts` covers both the Node and the browser build.

Closes #1057

*✍️ Drafted by Claude Code*

### DIFF
--- a/src/apify_api_error.ts
+++ b/src/apify_api_error.ts
@@ -8,15 +8,32 @@ import { isBuffer } from './utils.js';
 export type { ApifyApiErrorType } from './models.js';
 
 /**
- * Examples of capturing groups for "...at ActorCollectionClient._list (/Users/..."
- * 0: "at ActorCollectionClient._list ("
+ * Examples of capturing groups for "...at ActorCollectionClient.listResources (/Users/..."
+ * 0: "at ActorCollectionClient.listResources ("
  * 1: undefined
  * 2: "ActorCollectionClient"
  * 3: undefined
- * 4: "list"
+ * 4: "listResources"
  * @private
  */
-const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\._?([A-Za-z]+) \(/;
+const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\.([A-Za-z]+) \(/;
+
+/**
+ * A public method that returns the promise of a shared helper without awaiting it leaves only the helper on
+ * the stack, so the helper is reported under the method the caller invoked.
+ * @private
+ */
+const PUBLIC_METHOD_BY_HELPER: Record<string, string> = {
+    getResource: 'get',
+    updateResource: 'update',
+    deleteResource: 'delete',
+    waitForJobFinish: 'waitForFinish',
+    listResources: 'list',
+    createResource: 'create',
+    getOrCreateResource: 'getOrCreate',
+    addRequestBatch: 'batchAddRequests',
+    addRequestBatchWithRetries: 'batchAddRequests',
+};
 
 /**
  * An `ApifyApiError` is thrown for successful HTTP requests that reach the API,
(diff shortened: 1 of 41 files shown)


src/apify_api_error.ts
```diff
@@ -8,15 +8,32 @@ import { isBuffer } from './utils.js';
...
-const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\._?([A-Za-z]+) \(/;
+const CLIENT_METHOD_REGEX = /at( async)? ([A-Za-z]+(Collection)?Client)\.([A-Za-z]+) \(/;
+
+/**
+ * A public method that returns the promise of a shared helper without awaiting it leaves only the helper on
+ * the stack, so the helper is reported under the method the caller invoked.
+ * @private
+ */
+const PUBLIC_METHOD_BY_HELPER: Record<string, string> = {
+    getResource: 'get',
```
janbuchar: I'm not sure I understand why this is necessary

janbuchar (commented):
Looks reasonable to me, nice job! Maybe we should also make private methods and attributes #private while you're at it?

src/apify_api_error.ts
vdusek: Whoops, it landed on v3 already through #1046. It's there only to map the helper frame (`getResource`) back to the public method (`get`) - for logs and stack traces (?). Probably not necessary and removing it.

janbuchar (approved):
LGTM!

PR #1060, opened 2026-09-13
fix!: convert date fields through the zod schemas instead of the *At suffix heuristic

### LINKED ISSUE #522: `parseDateFields` skips date fields nested deeper than 3 levels

vdusek:
### Issue

Closes #522

### Description

`parseDateFields()` walked every response and turned any field whose name ends in `At` into a `Date`, three levels deep. Anything deeper stayed a string, and dates inside `userData`, task `input` and other caller-owned blobs were converted whether the caller wanted it or not. Neither behavior was documented.

The zod schemas do the conversion now, as pydantic does in the Python client. The emitter turns `format: date-time` into `z.iso.datetime({ offset: true }).pipe(z.coerce.date())`, so a field the specification declares as a date-time comes back as a `Date` at any depth and nothing else is touched. The ISO check runs ahead of the coercion because `new Date(null)` is epoch 0. `spec_guards.ts` compares the generated types against the schemas' `z.output`, which closes a latent hole as well, since `parseResponse()` hands back the output. `src/generated/schemas.ts` was edited in place rather than regenerated because the live specification has drifted, but a fresh emitter run over the recorded version reproduces it byte for byte.

### Breaking changes

- Date strings inside bodies the API stores verbatim (`userData`, task `input`, ...) are no longer converted.
- A `date-time` field carrying anything other than an ISO 8601 date-time with a `Z` or a time-zone offset throws `ResponseValidationError` where it used to pass through as a string.
- `RequestQueue.expireAt` is removed from the published type. The API never returned it.
- `parseResponse()` lost its `shouldParseField` parameter and `parseDateFields()` is removed. Both were internal.

The v3 upgrading guide covers the user-facing part.

*✍️ Drafted by Claude Code*


src/utils.ts
```diff
@@ -157,59 +153,6 @@ export function catchNotFoundForResourceOrThrow(err: ApifyApiError, resourceId:
...
- * Traverses JSON structure and converts fields that end with "At" to a Date object (fields such as "modifiedAt" or
- * "createdAt").
- *
- * If you want parse other fields as well, you can provide a custom matcher function shouldParseField(). This
- * admittedly awkward approach allows this function to be reused for various purposes without introducing potential
- * breaking changes.
- *
- * If the field cannot be converted to Date, it is left as is.
- */
-export function parseDateFields(
```
janbuchar, line 174: 🎉

janbuchar (approved):
Awesome, thanks!

Issue #1064, opened 2026-09-14
Make request-body compression configurable, mirroring the Python client


vdusek:
Compression is hard-wired in `maybeCompressValue()` (`src/utils.ts`): brotli at quality 6 with a silent gzip fallback, and no way to pick the algorithm, tune the quality, or plug in a custom compressor.

The Python client has an `HttpCompressor` base class with `GzipHttpCompressor` / `BrotliHttpCompressor` and a `compression` option on `ApifyClient` (apify/apify-client-python#927, [`http_compressors`](https://github.com/apify/apify-client-python/tree/master/src/apify_client/http_compressors)). Port that design.

*✍️ Drafted by Claude Code*

vdusek:
resolved by https://github.com/apify/apify-client-js/pull/1065

PR #1065, opened 2026-09-14
feat!: make request-body compression configurable

vdusek:
### TLDR

- Ports the `http_compressors` design of the Python client (apify/apify-client-python#927) to the JS client.
- Closes: #1064

### Description

Request compression was hard-wired in `maybeCompressValue()`: brotli at quality 6, a silent fallback to gzip, and no way to pick the algorithm, tune the quality, or plug in a custom compressor.

The client now has an abstract `HttpCompressor` (`contentEncoding` plus an async `compress()`), a `GzipHttpCompressor` (quality 1 to 9, default 9) and a `BrotliHttpCompressor` (quality 0 to 11, default 6), all exported from the package. `ApifyClient` takes a `compression` option: `'brotli'` (the default), `'gzip'`, or an `HttpCompressor` instance. The option is validated by shape, so an instance from a second copy of the package passes too. `HttpClient` holds the resolved compressor and the compress interceptor uses it. The rules for which bodies get compressed are unchanged: at least 1 KiB, a compressible content type, no caller-set `Content-Encoding`, Node.js only.

Two deliberate differences from Python. Brotli stays the default, since Node ships it in `node:zlib` and the dependency reason for Python's gzip default doesn't apply. Streams are never compressed, where Python reads file-likes into memory. `setRecord()` still has no `contentEncoding` option, so the docs route pre-compressed bodies through `client.httpClient.call()`.

Docs: a new `HTTP compression` concepts page adapted from the Python one, a section in the v3 upgrade guide, and the regenerated api-extractor report.

### Breaking change

The silent fallback from #990 is gone. The configured compressor is the only one the client runs, and its error fails the request, as in Python. A runtime whose `node:zlib` lacks brotli used to get a gzipped or uncompressed body and now needs `compression: 'gzip'`. A client constructed without the option sends the same requests as before.

*✍️ Drafted by Claude Code*


docs/02_concepts/07_http-compression.md
janbuchar, line 1: I'm still not convinced that this matters enough to warrant a fully-fledged guide.

src/http_compressors/base.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ *     async compress(data) {
+ *         return data;
+ *     }
+ * }
+ *
+ * const client = new ApifyClient({ token: 'my-token', compression: new IdentityCompressor() });
+ * ```
+ */
+export abstract class HttpCompressor {
```
janbuchar, line 26: Abstract class with abstract methods only sounds like an interface.

janbuchar (approved):
Pre-approving, thanks

src/http_compressors/base.ts
vladfrangu: Not really, `instanceof HttpCompressor` works but you can't check with code if something implements an interface (without zod et al)

janbuchar: True that, but we don't really do `instanceof` much here, do we? I'd say that pitfalls such as breaking if you have multiple versions of `apify-client` in your `node_modules` or when crossing realms (service workers) still apply here. Typescript + a rough-shape-check via `zod` is preferable IMO.

docs/02_concepts/07_http-compression.md
vdusek: let's review the whole docs as part of https://github.com/apify/apify-client-js/issues/851 and align it with the Python version

src/http_compressors/base.ts
vdusek: changed to interface

PR #1066, opened 2026-09-14
fix!: narrow `pushItems()` to the item shapes the API actually accepts

barjin:
The dataset items endpoint only ever accepts an object or an array of objects ([source code](https://github.com/apify/apify-core/blob/b13e101720f6fbbda842bcfd61e0324ce2eaf3c4/src/api/src/routes/datasets/items.ts#L109)), but `pushItems()` typed and validated a plain string array as acceptable input. This always fails at request time.

Passing `string` only works if the `string` is a JSON-serialized representation of the other two types.


src/resource_clients/dataset.ts
```diff
@@ -305,13 +305,14 @@ export class DatasetClient<
...
      * Stores one or more items into the dataset.
      *
-     * Items can be objects, strings, or arrays thereof. Each item will be stored as a separate
-     * record in the dataset. Objects are automatically serialized to JSON. If you provide an array,
-     * all items will be stored in order. This method is idempotent - calling it multiple times
-     * with the same data will not create duplicates, but will append items each time.
+     * Each item will be stored as a separate record in the dataset. Objects are automatically
+     * serialized to JSON. If you provide an array, all items will be stored in order. This method
+     * is idempotent - calling it multiple times with the same data will not create duplicates, but
+     * will append items each time.
```
barjin: > This method is idempotent - calling it multiple times with the same data will not create duplicates, but will append items each time.

I wonder what the original author meant by this, that doesn't sound very idempotent to me 😅

barjin, line 311: > This method is idempotent - calling it multiple times with the same data will not create duplicates, but will append items each time.

I wonder what the original author meant by this, that doesn't sound very idempotent to me 🤔

janbuchar: image

That's just a hallucination.

janbuchar:
Doesn't the `string` signature mean "pre-serialized" JSON? I.e., it is taken verbatim and sent with a `text/json` content type to the API?

barjin:
>Doesn't the string signature mean "pre-serialized" JSON? I.e., it is taken verbatim and sent with a text/json content type to the API?

image

I believe so. But I still think `string[]` is wrong either way (`axios` will serialize it, so we're sending a JSON array of strings, which causes the server-side validation to fail.

janbuchar (approved):
LGTM, sorry for misreading the intent of the PR

